PR #1, opened 2021-12-13
Fix M1 mac development experience

kdutia:
Replace psycopg2 with psycopg2-binary. See climatepolicyradar/policy-search#142

There is still the remaining question about running `poetry config virtualenvs.create false` in `make poetry_environment`. This also breaks the dev experience on my mac, and seems like the wrong behaviour - we want poetry to make a virtualenv when coding *outside* of Docker.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,9 +5,6 @@
 /app/.next/
 /app/out/
 
-# Poetry
-poetry.lock
-
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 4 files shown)


chrisaballard (approved):
Reviewed changes. I had some problems installing pycopg2-binary until I realised that the `poetry.lock` file needed to be updated.

This needs to be committed to the repo, so I have also added `poetry.lock` to make sure we use the latest dependency versions.

I have also updated the `Makefile` to ensure that a poetry virtual environment is set up when doing a local `dev_install`.

PR #4, opened 2021-12-17
Database migration to create initial database structure

chrisaballard:
I have added a database migration that will create the initial database structure.

Documentation for the structure is available [here](https://www.notion.so/climatepolicyradar/Database-schema-80dd628cfe484811a722d8b4d4a673ad).

To run the migration on your local dev database, run the following command:

```
alembic upgrade head
```

This will create all the tables and foreign key constraints.

Note: that I have not yet created a script to populate the following lookups:

- language
- geography
- source
- action_type

I will concentrate on adding this next

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,10 +100,16 @@ npm install
 npm test
 ```
 
-## Migrations
+## Database migrations
 
 Migrations are run using alembic. To run all migrations:
 
+Local dev environment:
+```
+alembic upgrade head
+```
+
+Docker dev environment:
 ```
 docker-compose run --rm backend alembic upgrade head
 ```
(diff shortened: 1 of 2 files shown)


kdutia:
@chrisaballard the command runs successfully and the ER diagram generated by PgAdmin looks to be the same as the one you designed, from visual inspection. Is there anything else I should check for before approving this?

kdutia (approved):

chrisaballard:
I have created a new database migration to populate standard lookup values for the following tables:

- language
- geography
- action_type
- source

Also made some of the columns on the `language` and `geography` tables nullable as those fields were not present in the lookup data.

The data is loaded from two text files which are stored in `/backend/app/alembic/versions/lookups`:

- `language-iso-639-3.txt` - iso 639-3 standard language codes
- `geography-iso-3166.csv` - contains country codes that were defined for the original prototype.

The downgrade migration will remove the inserted lookup values from the tables and will reset the columns make not not_nullable.

Run `alembic upgrade head` to run the migration

kdutia:
@chrisaballard where is the data stored which populates the source and action_type lookup tables?

PR #9, opened 2021-12-20
Feat 141: automatic unit tests

kdutia:
Github actions runs tests on the frontend (using cypress) and backend (using pytest) using the docker-compose stack, with the environment variables in `.env.example`. These tests are run on pushes or pull requests to the `main` or `dev` branches.

In future we might want to modify this work so that:

- environment variables are retrieved from our github secrets, if we have a hosted test environment
- `cypress run` or `pytest` commands run with flags or config of our choosing.

This PR will act as a test of this work 😉

Closes https://github.com/climatepolicyradar/policy-search/issues/141


chrisaballard (approved):
Reviewed - tested frontend and backend tests work locally. I could get the backend pytest tests working fine, but for some reason had a problem with the backend tests - doesn't seem to connect to the cypress docker container. However, both backend and frontend tests run fine as the github action.

I've made a small tweak to add a status batch to the readme. I did this to test the action was working ok, but also because it could be super helpful!

Some suggested improvements we could make later:
1. One improvement we could make would be outputing the pytest results in junit format which would then allow us to add a test status report displayable in github so that interpreting test results is easier.
2. Adding pytest-coverage so that we can view the coverage of unit tests - will help us to monitor how well tested our codebase is. Will add some issues to the backlog for these.

Overall though, this is all good and working as expected :)

PR #10, opened 2022-01-05
feat-140-db-structure - create standard lookup values

chrisaballard:
Created an alembic db migration to populate the following tables with lookup values:

- language
- geography
- action_type
- source

`language` and `geography` are populated from two files located in `/backend/app/alembic/versions/lookups`:

- `language-iso-639-3.txt`: standard iso-639-3 lookup codes
- `geography-iso-3166.csv`: country codes derived from iso 3166 that we used in the original prototype.

`action_type` is populated with the following values:

- Law
- Policy

`source` is populated with a single value:

- CCLW

Run `alembic upgrade head` to apply the migration.

### DIFF
--- a/backend/app/alembic/versions/9ef4aeb8094f_populate_lookups.py
+++ b/backend/app/alembic/versions/9ef4aeb8094f_populate_lookups.py
@@ -1,0 +1,128 @@
+"""Populate lookups
+
+Revision ID: 9ef4aeb8094f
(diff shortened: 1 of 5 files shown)


kdutia (approved):
Ran the migration and `SELECT * FROM <table>` queries for each of these lookup tables. Results as expected 👍🏼

Issue #13, opened 2022-01-06
API > language, geography, source and action type lookups 


kdutia:
e.g. `GET /api/v1/language`

We could potentially do this in a generic way for any lookup table.

Issue #11, opened 2022-01-06
API > Submit a new policy


kdutia:
@chrisaballard commented on [Tue Dec 07 2021](https://github.com/climatepolicyradar/policy-search/issues/136)

Create a REST api that a new policy can be submitted to. This should support a POST request.

- [x] Define structure of POST request
- [x] Create API
  - [x] document upload API
  - [x] action submission API
- [x] Write tests (including mocking s3)
- [x] Document API using OPENAPI json - this should be visible using the fastapi documentation support
- [x] Return a response indicating that policy has been submitted, or an error has occurred

---

@kdutia commented on [Thu Dec 16 2021](https://github.com/climatepolicyradar/policy-search/issues/136#issuecomment-995609383)

@chrisaballard @eurolife here's a draft of the POST request structure (first point above) for your review https://www.notion.so/climatepolicyradar/Policy-submission-API-design-585004266c064fa8b0d5cb9e8bbba6ce

eurolife:
@kdutia 
Issues/comments on the api:

- Can’t send null values for month/day so sending 1 and 1 until backend can handle null values.
- Looks like there is a file size limit, but not sure what it is or where it’s set. 
- Submit document error - looks like it could be issue with s3 and credentials (error log below)
- fyi: I changed ACCESS_TOKEN_EXPIRE_MINUTES to 300 (security.py) instead of 30 so I don’t have to keep renewing it - not pushing this change though in case we need to keep it at 30.

`INFO:     172.19.0.5:36244 - "POST /api/v1/document HTTP/1.0" 500 Internal Server Error

ERROR:    Exception in ASGI application

Traceback (most recent call last):

  File "/usr/local/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 389, in run_asgi

    result = await app(self.scope, self.receive, self.send)

  File "/usr/local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 45, in __call__

    return await self.app(scope, receive, send)

  File "/usr/local/lib/python3.8/site-packages/fastapi/applications.py", line 208, in __call__

    await super().__call__(scope, receive, send)

  File "/usr/local/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__

    await self.middleware_stack(scope, receive, send)

  File "/usr/local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__

    raise exc

[...]

PR #24, opened 2022-01-07
Create action and document submission APIs; lookup APIs

kdutia:
This PR contains:
* `POST /action` and `POST /document` APIs to submit actions and upload documents
* `GET /geographies`, `GET /languages`, `GET /action_types`, `GET /sources` APIs to get IDs and metadata for each of these tables.

Linked to #11 and #13.

I've created this PR to not hold up frontend work, **however the following work is still needed**:

1. There's an issue with the length of the `name` field in the reponse to `GET /sources`. It looks like this is because the database column is of type `Character(1024)` rather than `VARCHAR()`. I'll fix this in a separate PR, as the load script is already on the `dev` branch.
2. The test database isn't picking up the latest lookup table values, meaning that in test, each lookup API returns an empty object. This may just be worth raising a bug issue for next sprint - @chrisaballard would appreciate your thoughts when we review this code.


eurolife (approved):

kdutia:
@chrisaballard although this PR is merged it'd still be helpful if you could review my code please. Point 2 in the PR description around the test database is something we should look to fix in a future sprint, too.

PR #26, opened 2022-01-10
Add alembic revision to change dtype of 'name' col in 'source' table

kdutia:
Added an alembic revision to change the data type of the `name` column in the `source` table from *character* to *character varying (128)*. This means that trailing spaces are now not stored in this column.

Inspecting the table contents in pgAdmin seems to show that now only the 4 characters of 'CCLW' are stored for the source we have in there at the moment.


chrisaballard:
Ran migration and column was altered without issue.

chrisaballard (approved):
Approved pr

PR #30, opened 2022-01-12
Feat 11 - policy submit APIs - second set of changes

kdutia:
In this PR:
* Allow `POST /document` and `POST /action` APIs to take `null` values for `day` and `month` fields. These values are then stored as `1` in the database.
* Change nginx config to allow file uploads up to 150MB.
* Add tests for 400 errors in `POST /action` API.
* Filter languages returned by the `GET /languages` API to only those with a part1_code - now 184 languages.
* Fix wrong document URL being added to document DB; `POST /action` API now returns correct (new) URL of each document

@eurolife please could you check these changes solve the issues you raised in a comment on #11?
@chrisaballard please could you review this branch as a whole?


eurolife (approved):

eurolife:
@kdutia s3_url stored in db still not quite right (still has None). Example:

https://cpr-document-store.s3.None.amazonaws.com/user-1-time-20220112194536-Tracked_Returns_label_QX309977193GB.pdf

kdutia:
@eurolife as per our chat this morning this was due to a missing environment variable. After [this commit](https://github.com/climatepolicyradar/navigator/commit/4f6363cefe365f2d8333ed484f29cf74845864a0), the backend now raises an appropriate error upon startup if there are any missing environment variables.

chrisaballard (approved):
Reviewed implementation of API and associated unit tests. I think a great job has been done of defining the api routes for lookups and posting actions and documents. Unit tests cover the essential aspects of verifying that post requests to the document and action routes is functioning ok. A few observations, but nothing that prevents merging this PR:

- We should make sure that full docstrings are written for all code - currently only very summary level docstrings are in place. I've added a techical story to the backlog for this.
- I think we should create smaller unit tests each with a discrete focus. At the moment I think that the `test_post_action` unit test is quite large and tests multiple elements. These should be broken down into discrete individual tests.

Approving PR.

chrisaballard:
@kdutia thanks for merging! :)

PR #33, opened 2022-01-17
Feature/122 submit new policy

eurolife:
User should be able to create a new action and add multiple document files or document urls to it. This is submitted to the db via the api.


kdutia:
@eurolife the frontend on my machine is making calls to incorrect URLs for the lookup APIs, e.g. `GET /undefined/sources`. 

It seems from your code like I'm missing an `NEXT_PUBLIC_API_URL` variable? What should the value of this be, and can I put it in the main `.env` as your code's written or does it need to go somewhere else?

If possible could it go in the main repo `.env`, and please could you update `.env.example` and `.env.template`?

eurolife:
> @eurolife the frontend on my machine is making calls to incorrect URLs for the lookup APIs, e.g. `GET /undefined/sources`.
> 
> It seems from your code like I'm missing an `NEXT_PUBLIC_API_URL` variable? What should the value of this be, and can I put it in the main `.env` as your code's written or does it need to go somewhere else?
> 
> If possible could it go in the main repo `.env`, and please could you update `.env.example` and `.env.template`?
> 
> Also, the heading in the page says "Submit new action" and the title in the tab says "Submit new policy"

@kdutia Sorry completely forgot about that. I have added a .env.example file with the url value. Add this to a .env or .env.local file. Unfortunately I'm not aware of any way for the frontend to read the .env from the root folder so this must reside inside the frontend folder. The heading title has been corrected in my other branch where I have implemented the corporate branding.

PR #71, opened 2022-02-03
Node dev docker

opyate:
A more fluid local docker-based dev approach for the frontend:
- node_modules inside and outside container doesn't clash
- live-reload inside container when code changes outside container
- a documented way to add new packages to the running container
- all without requiring rebuilds


opyate:
I'll look at the test failure (permission issue).

PR #339, opened 2022-02-08
refactor docs

opyate:
The current documentation clumps together docker-based + local development approaches, so the rewritten docs separates these.
Also, the Makefile was broken up to reflect this.


chrisaballard:
Reviewed - this all looks great, much clearer than it was previously.

eurolife (approved):

PR #342, opened 2022-02-10
Feat 41 test database schema

opyate:
All Python devs, your DB volume will have an incompatible `alembic_version` after merging this, so please either:
- delete volumes with `docker-compose down -v` (recommended)
- or, if you want to keep your data, add this row to `alembic_version`:

```
navigator=# select * from alembic_version;
 version_num 
-------------
 0004
```


chrisaballard:
This all looks good. I think the numeric revision ids will make it much clearer to see the migration order.

I'm a little confused by the purpose of `generate-sql.sh` though. @opyate can you clarify this?

Regarding the test in `test_actions.py` that adds records to Geography, ActionType etc. Should this be moved into a different test as `test_post_action` relates to testing a post to the api?

opyate:
> I'm a little confused by the purpose of `generate-sql.sh` though. @opyate can you clarify this?

When you run it, it created "up" and "down" SQL:

```
docker-compose run --rm backend alembic/generate-sql.sh
```

Output, up:

```
BEGIN;
-- Running upgrade 0003 -> 0004

ALTER TABLE source ALTER COLUMN name TYPE VARCHAR(128);

UPDATE alembic_version SET version_num='0004' WHERE alembic_version.version_num = '0003';

COMMIT;
```

Output, down:

```
BEGIN;
-- Running downgrade 0004 -> 0003

UPDATE alembic_version SET version_num='0003' WHERE alembic_version.version_num = '0004';

COMMIT;
```

This allows us to upgrade the DB by hand, and downgrade if need be.


> Regarding the test in `test_actions.py` that adds records to Geography, ActionType etc. Should this be moved into a different test as `test_post_action` relates to testing a post to the api?

With the newly-added `ForeignKey` declarations in the model, this test fails without these records.

chrisaballard:
Thanks for the explanation @opyate that makes sense now.

PR #343, opened 2022-02-13
Feature/51 action search UI

eurolife:
Non-functional search start page added which can be accessed at http://localhost:3000
- User can enter a search term 
- When a search term is entered, an 'X' button is visible to allow the user to clear the search input if needed
- User can click the 'X' button to clear search input
- Pressing enter or the 'Go' button will trigger a search but at this time, this is not functional.
- This page is responsive and works on all screen sizes.

FYI the 'Add Action' form is still available but moved to http://localhost:3000/add-action 

TODO: Write unit tests for this new page.


kdutia:
@eurolife got it up and running and looks great! The dashed outline that appears in the search box when it's in focus seems like an unusual design decision, though. I think it's redundant because the white search box is really clear on the dark background.

It also means the dashed box resizes as the cross to remove search terms appears and disappears which is a bit distracting.

Dashed outline on search bar focus
image

Outline resizes when user starts typing and cross appears (and resizes again if I clear my search)
image

kdutia:
@eurolife after sprint review we agreed on the above dotted lines comment and identified two issues, which I think can be tackled in future issues rather than as part of this PR:

- The 'Search for something...' text gets cut off on mobile view. Here it'd be great if we could just have the example.
- There is a vertical line in the background which looks like an unintended artefact (img 1).
- It'd be great if the blocks showing the numbers of documents, jurisdictions and new documents could be a bit more opaque to make them clearer to read over the background on small screens (img 2).

Img 1
image

Img 2
image

Approving on the basis that the dotted lines are removed and issues for these other tasks are raised 🎉

kdutia (approved):
See comments

eurolife:
@kdutia I've made the changes noted above. Let me know if you'd like to check it again before merging.

Issue #349, opened 2022-02-14
Technical > Create unit tests and update github actions for testing pipeline


chrisaballard:
Create unit tests for the new pipeline pdf2text cli.
Add github action which can set up a pipeline docker container in which to run the pipeline tests. This will ensure that pdfalto and other runtime dependencies are correctly installed when running the tests.

chrisaballard:
Tested changes in PR https://github.com/climatepolicyradar/navigator/pull/361 - all works ok (see comments in PR)

PR #350, opened 2022-02-14
pdf2text cli to process pdfs

chrisaballard:
I have added a new pipeline folder to the repo that contains the beginnings of a batch process to extract text from pdf documents.

The initial implementation is as a cli which can be used to extract the text, text blocks and positional information from a set of pdf documents. This uses the pdfalto parser (used by grobid) to extract the text embedded in each document.

A Dockerfile is provided which can be used install the pdfalto dependencies and run the pipeline. Usage of the cli and instructions on how to build the docker image are described in `/pipeline/README.md`.

Note: currently no unit tests have been defined. This is because we will need to create a new github action to run the pipeline tests in the pipeline docker container. There was insufficient time in this sprint to do that, so have raised a new issue #349 for this.

### DIFF
--- a/pipeline/Dockerfile
+++ b/pipeline/Dockerfile
@@ -1,0 +1,19 @@
+FROM python:3.8
+
+RUN mkdir /app
+WORKDIR /app
+
+RUN apt update && \
+    apt install -y cmake
+
+RUN mkdir ~/.ssh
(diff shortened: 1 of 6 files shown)


kdutia (approved):
@chrisaballard tested & looks good to me. Sorry it took a while to review - unfortunately I can't get it running locally due to M1 issues so had to run it in an EC2 machine.

PR #359, opened 2022-02-18
Feat 349 parsing pipeline tests

kdutia:
Wrote:

* unit tests for methods for the `Document` class
* a test to ensure that `DocumentEmbeddedTextExtractor` returns a `Document` with the correct filename and nonempty `text_blocks`

These tests run and pass on my machine as long as the `PDFALTO_PATH` env variable is set.

The work to get these tests running in github actions is still outstanding for this issue.


opyate:
> These tests run and pass on my machine as long as the `PDFALTO_PATH` env variable is set.

Worth putting it in `.env*`?

pipeline/test/test_extract.py
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+
+def test_document_to_string(document):
+
+    assert document.to_string() == SEPARATOR + "paragraph 1" + "\n"
+
+
+def test_document_save_text(document, tmpdir):
+    output_path = os.path.join(tmpdir, "test_document.txt")
```
opyate, line 71: I know it's only `/tmp`, but might still be worth cleaning up when the test finishes?

kdutia: It seems like pytest only keeps [max 3 temporary folders](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-default-base-temporary-directory), and I'm thinking it might be useful to keep these around in case a test fails?

What's the advantage of cleaning up after the test finishes/is it best practice?

kdutia:
> > These tests run and pass on my machine as long as the `PDFALTO_PATH` env variable is set.
> 
> Worth putting it in `.env*`?

It's in my local `.env` - not sure what its value will be for `.env.example` as this will depend on the Github actions setup which I'll add to the next PR

pipeline/test/test_extract.py
opyate: Ah, please ignore me. Just realised it's the pytest `tmpdir` fixture.

PR #361, opened 2022-02-18
Github actions integration for parsing pipeline tests

### LINKED ISSUE #349: Technical > Create unit tests and update github actions for testing pipeline

kdutia:
Integrated unit tests for the PDF parsing pipeline into the Github Actions CI process. Closes #349.

I had some issues enabling SSH access from Github actions to Github so `git submodule update` could be run, so instead changed the Dockerfile to use the HTTPS URL for the `xpdf-4.03` submodule. This also means that there is no buildkit dependency for Docker, and the readme has become a lot simpler.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,26 +11,26 @@ on:
       - dev
 
 jobs:
-  # test:
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #     - uses: actions/checkout@v2
-  #     - name: Build docker-compose stack using .env.example
-  #       run: cp .env.example .env && make build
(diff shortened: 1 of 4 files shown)


chrisaballard:
- Rebuilt the docker container using the new Dockerfile definition - builds ok.
- Tested running the cli inside the container - all works ok

Great idea @kdutia to override the url for the git submodule - it has made the process much simpler 👍

PR #362, opened 2022-02-21
Feat-346-extract-nested-pages

chrisaballard:
The pdf2text text extraction pipeline has been updated to return a nested structure for a document. The text blocks extracted in each document are nested by the page in which they occur. The dimensions for each page are now also returned. The coordinates of each TextBlock are now stored as a list of 2-element float tuples. This ensures that more than 4 coordinates can be used for each text block in future - allowing non-rectangular text block regions.

In addition, the text in each text block is now returned as a list of lines, instead of a concatenated string. This will make it much easier to post-process the text blocks. E.g. word hyphenation can be detected by looking for lines ending with a hyphen.

The following are the changes to the main classes storing the document structure:

```
class Document:
   pages: List[Page]
   ...
```

```
class Page:
   text_blocks: List[TextBlock]
   dimensions: Tuple[float, float]
   ...
```

```
class TextBlock:
   text: List[str]
   ...
   coords: List[Tuple[float, float]]
   ...
```


chrisaballard:
Note: that this PR does not include the new definition for the pipeline Dockerfile covered in PR https://github.com/climatepolicyradar/navigator/pull/361

kdutia (approved):
Checked code, ran CLI and ran tests locally (as they're not yet in GH actions). Looks great 👍🏼

Issue #50, opened 2022-02-01
Navigator documents should be populated with policy documents from CCLW


chrisaballard:

chrisaballard:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Tech-Team-613867967a1a9200123c9aad/issues/climatepolicyradar/navigator/50?planning-poker) with ZenHub @eurolife @kdutia @opyate

opyate:
Going into next sprint: still need to implement tests.

Issue #49, opened 2022-02-01
Database > Navigator actions should be populated with policies from CCLW


chrisaballard:
<html>
<body>
<p>Navigator action documents should be initially populated from CCLW.</p>
<p>Use <a href="https://drive.google.com/drive/u/0/folders/19E9oEP-6bswBlUTwIlzQpcQ44RzCe5Zs">CCLW export files</a> to map and load CCLW data into Navigator.</p>
<p>The data should be mapped to Documents using the following mapping.</p>

CPR | CCLW
-- | --
Document name | Document name
Document source | CCLW
Document description | Description
Document date | Policy date
Document class | Document type
Document source url | CCLW URL to document
Document language | [link] [link]

<p>Each document should be linked to the associated Action for that policy in Navigator.</p>
<p>The documents should be filtered such that only documents with a mime type of PDF are loaded.</p>
<p>Note: Other document mime types will not be loaded for the alpha.</p>

</body>
</html>

From prototype:
* [this notebook](https://github.com/climatepolicyradar/cpr/blob/master/nbs/policy-etl.ipynb) transforms CCLW data to a processed_policies csv file 
* [this CLI](https://github.com/climatepolicyradar/policy-search/blob/dev/cli.py) loads the processed_policies csv file into the prototype 

Some work has been done on action and document submission in #24

chrisaballard:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Tech-Team-613867967a1a9200123c9aad/issues/climatepolicyradar/navigator/49?planning-poker) with ZenHub @eurolife @kdutia @opyate

PR #367, opened 2022-02-22
Feat 49 populate actions from cclw

opyate:
First pass of populating actions from CCLW.
- uses hard-coded CSVs
- uses code from prototype (pandas, etc)
- rudimentary tests


chrisaballard:
@opyate I've reviewed the changes, please find my observations below.

## Flattening of policies and documents

In the original prototype, we flattened out the CCLW data, and `loader` is working the same way. This means that currently, an Action is being created for each unique policy and document combination. Instead, it should create a unique Action for each CCLW policy. All documents associated with that policy should then be stored as Documents against the Action ([[issue #50](https://app.zenhub.com/workspaces/climate-policy-radar-613867967a1a9200123c9aad/issues/climatepolicyradar/navigator/50)](https://app.zenhub.com/workspaces/climate-policy-radar-613867967a1a9200123c9aad/issues/climatepolicyradar/navigator/50)).

## Using backend and front end export files

We’re currently using both the backend and front end export files. This was originally because the backend file includes CCLW policy ids, but does not include document URLs to documents stored on CCLW blob storage. These are available on the front end export file, and so the two are merged together. However, we’re not loading the source policy id into our database any more, so I don’t think its relevant to use the backend export file? In which case, we could simplify the process considerably and just use the existing front end export file and not worry about merging them.

## Flexible bulk load format

The current loader is very fixed to the current CCLW export file structure. This means it couldn’t be used to bulk load policies from other sources. We originally discussed the possibility of defining our own bulk load file schema which would enable this to happen. This would involve then mapping the CCLW data to our schema. I think the scope of this issue for the sprint can be limited to supporting CCLW only, but I think we should raise a new issue in the backlog to make sure that we consider this going forward.

## Other observations

- `extract.main.extract(...`  has hardcoded paths to data referenced by date. I think we would need to change this to allow the path and filenames to be specified as arguments to the `[main.py](http://main.py)` script. This will enable us to repeat the import process for an updated set of import files, without needing to update the code.
- Check fully qualified import in `extract.main` - this refers to `app.model` does this work ok?
- `transform.datafixes` updates the dates for 5 policies which have missing dates. I guess its possible that in an updated extract that new policies could be received which also have missing dates. Would it be better to update the source data and instead include validation in the process to warn the user that some policies (list them) won’t be loaded because they don’t have date information?
- As part of [[issue #50](https://app.zenhub.com/workspaces/climate-policy-radar-613867967a1a9200123c9aad/issues/climatepolicyradar/navigator/50)](https://app.zenhub.com/workspaces/climate-policy-radar-613867967a1a9200123c9aad/issues/climatepolicyradar/navigator/50), we’ll need to decide how to deal with document names. Currently these have very uninformative names, and I would recommend that they are manually reviewed by @danny-waite-cpr and @marcusodavies  prior to loading.
- The load will currently bulk load policies in irrespective of whether the load has been carried out previously and would create duplicate Actions. I think for the scope of this sprint that is fine, but we should capture this in the backlog to make sure that we can take it into account. This is where storing source policy ids may be helpful to update them.

opyate:
@chrisaballard thanks for the comments!

> In the original prototype, we flattened out the CCLW data, and `loader` is working the same way. This means that currently, an Action is being created for each unique policy and document combination. Instead, it should create a unique Action for each CCLW policy. 

Definitely. When I implement https://github.com/climatepolicyradar/navigator/issues/50 next, I'll aggregate those docs into a single action. (In the current code, the dupe actions are blocked with the new unique constraint.)

> We’re currently using both the backend and front end export files. This was originally because the backend file includes CCLW policy ids, but does not include document URLs to documents stored on CCLW blob storage. These are available on the front end export file, and so the two are merged together. However, we’re not loading the source policy id into our database any more, so I don’t think its relevant to use the backend export file? In which case, we could simplify the process considerably and just use the existing front end export file and not worry about merging them.

Shall I simplify it in a subsequent task? 

>     * `extract.main.extract(...`  has hardcoded paths to data referenced by date. I think we would need to change this to allow the path and filenames to be specified as arguments to the `[main.py](http://main.py)` script. This will enable us to repeat the import process for an updated set of import files, without needing to update the code.

I'll make that change now.

>     * Check fully qualified import in `extract.main` - this refers to `app.model` does this work ok?

It's loader's own `app.model`, not the one from backend.

>     * `transform.datafixes` updates the dates for 5 policies which have missing dates. I guess its possible that in an updated extract that new policies could be received which also have missing dates. Would it be better to update the source data and instead include validation in the process to warn the user that some policies (list them) won’t be loaded because they don’t have date information?

This just came up in a call with Danny and Marcus, actually. We'll implement a warning/flagging system, then they can fix the CSV and re-upload.

chrisaballard:
Reviewed recent changes. All looks fine - removing the backend export files from `extract` and `load` has simplified the load process considerably. I've just notived that issue #49 seems to refer to documents and not actions, apologies if there was any confusion there!

PR #370, opened 2022-02-23
Feature/229 localisation

eurolife:
Navigator is now set up with multi-lingual support. To test this:

- Make sure to run `docker-compose build --no-cache frontend` to add newly added dependencies
- Go to http://localhost:3000 to see the expected search start page
- Then go to http://localhost:3000/fr to see the same page but with text in French (ignore bad translations, I used Google translate)
- FYI text is now loaded from public/locales/[lang-code]/[namespace].json
- Multi-lingual support has also been added to the Add Action form.


opyate:
Left a comment out-of-band about possibly using natural keys.

Untested suggestion: Language JSON Before:

```
{
    "url": "Enter URL"
}
```

After:

```
{
    ["Enter URL"]: "Enter URL"
}
```

Then the code can be

```
                label={t('Enter URL')}
```

...which might give the code more context too.

Another benefit: a translator can look at a translation file in isolation, without having to look at the English equivalent side-by-side.

opyate (approved):

PR #371, opened 2022-02-24
Feat-345-extract-s3-documents

chrisaballard:
The pdf2text cli has been updated to support the processing of pdf documents stored on s3. This can now be used to batch extract text from documents in an s3 bucket, and then upload the extracted json and text files to a second bucket.

There have been some changes to the cli arguments. You no longer need to specify `--text` or `--json` arguments as this is implicitly assumed. There is a new argument `s3` which is used to inform the cli that you wish to process files on s3 rather than locally, and it will interpret the paths as s3 bucket/folder locations.

In this regard, the bucket folder locations are specified in the following format:

`[bucket]/[folder]/[folder]/...`

It will parse the string to obtain the bucket name and folder path and will read/write files into the corresponding bucket and nested folders.

In addition to this, the codebase has been refactored to introduce a common navigator package in `common/navigator`. This is now referenced in `backend` and `pipeline` and is included in the respective Dockerfiles. This means that any common code can be placed in this package and imported like so:

`from navigator.core.aws import S3Client`

I have not updated the unit tests for pipeline as all changes have been made to the `pdf2text` cli script rather than package/module code. I think later we may well want to put this code into separate modules, at which point unit tests could be created. But I think its worthwhile incorporating the Adobe API changes first to see what impact this has.


common/navigator/core/aws.py
```diff
@@ -150,6 +151,67 @@ def move_document(
             new_bucket, os.getenv("AWS_REGION"), new_key or s3_document.key
         )
 
+    def list_files(self, bucket: str, max_keys=1000) -> t.Union[S3Document, bool]:
```
kdutia, line 154: @chrisaballard should the return of this method be typed as [typing.Generator](https://docs.python.org/3/library/typing.html#typing.Generator)? I think something like `t.Union[t.Generator[S3Document, None, None], bool]`

common/navigator/core/aws.py
```diff
@@ -150,6 +151,67 @@ def move_document(
...
+        """
+
+        is_truncated = True
+        next_continuation_token = None
+        try:
+            while is_truncated:
+                # Include a continuation token in the arguments to boto3 list_objects_v2
+                # if we want to continue listing files from the last call
+
+                kwargs = {"Bucket": bucket, "MaxKeys": max_keys}
```
kdutia, line 175: My interpretation of the `max_keys` kwarg from the docstring was that it would limit the number of returned documents: I interpreted "maximum number of files to return in *a single call*" as being a single call of the `list_files` method. 

From looking at the code, it seems like it essentially controls the page size of the `list_files` method. 

Could you please update the docstring to clarify this? 

More generally, I think using 'a single call' to refer to a call of a function within the method being documented is probably bad practice, as it requires understanding the implementation of the function.

kdutia:
@chrisaballard I've added a couple of comments on code documentation above. Also, did you mean to delete `backend/poetry.lock`? I suggest we keep updating and committing our lock files.

chrisaballard:
Thanks @kalyan - I've made the changes you requested

common/navigator/core/aws.py
chrisaballard: Change made

Issue #357, opened 2022-02-15
Post-processing of extracted text > subscripts & superscripts


kdutia:
Note: for the prototype, the PDF parser was able to identify superscript and subscript text which was then ignored.

kdutia:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Tech-Team-613867967a1a9200123c9aad/issues/climatepolicyradar/navigator/357?planning-poker) with ZenHub @opyate @chrisaballard

kdutia:
@stefl14 my notebook with a WIP effort for this task is here. We can go through it this afternoon if you'd like https://github.com/climatepolicyradar/navigator/blob/feat-335-adobe-extract-pipeline/pipeline/temp-adobe-postprocess-sup-sub.ipynb

PR #373, opened 2022-02-25
Feat 335 adobe extract pipeline

kdutia:
In this PR:
* refactored extractors to separate out the 'PDF to data' and 'data to document' steps. This would allow us to cache the first step, which might be a costed API call, and separates out both steps for implementation in a pipeline.
* implemented an `AdobeAPIExtractor` class which extracts text from a PDF in natural reading order
* added `AdobeAPIExtractor` to the pdf2text CLI. When Adobe extraction fails, the CLI indicates this and falls back to using PDFalto.

In order to add the adobe PDF services SDK package I also refactored the Dockerfile to install poetry using the [default installer](https://python-poetry.org/docs/#osx--linux--bashonwindows-install-instructions) rather than pip. This means that if we have to `pip install` a package in the future in the Dockerfile because it won't install in poetry, any dependencies of this new package won't overwrite poetry's.

I'm going to start looking at unit tests for `AdobeAPIExtractor` this afternoon and I'll add them to this PR when done, or a new one if this is already merged.


kdutia:
I've added some unit tests for the PDF splitter and `AdobeAPIExtractor`. The one test that's remaining is one for when the Adobe API raises a 'file exceeds page limit' error.

pipeline/test/test_extract.py
```diff
@@ -139,3 +152,76 @@ def test_embedded_text_extractor(test_pdf_path):
...
+            test_pdf_path,
+            data_output_dir=data_output_dir,
+            output_folder_pdf_splits=split_dir,
+        )
+
+        assert doc.filename == test_pdf_path.name
+
+        # The returned document can have up to 8 pages (the number of pages in the original PDF).
+        # Pages with no parsed content (e.g. all figures) won't be added to the document.
+        assert len(doc.pages) <= 8
```
opyate, line 184: For this test, can the number of pages be verified exactly?

kdutia: Alternatively I can rewrite the code so that when the Adobe API doesn't detect any text on a page, a `Page` object is still returned with `TextBlocks==[]`. Maybe that would be better in general, as then `len(doc.pages)` would always equal the number of pages in the PDF?

kdutia: I don't think a page without text blocks is useful in itself, but it'd mean that 

- `Document` has a method we can call (`len(doc.pages)`) that gives us the length of the original PDF
- we can iterate through `[page in doc.pages]` and this would match up to the pages in the existing PDF.

@chrisaballard would be interested to hear your thoughts as part of your review

chrisaballard:
@kalyan should the following two files have been committed?

- `\pipeline\temp-adobe-postprocess-sup-sub.ipynb`
- `\pipeline\temp-adobe.json`

kdutia:
> @kalyan should the following two files have been committed?
> 
>     * `\pipeline\temp-adobe-postprocess-sup-sub.ipynb`
> 
>     * `\pipeline\temp-adobe.json`

They're WIP (the notebook uses the JSON) for #357, and artefacts of me having to move between EC2 and my local machine. I suggest we leave them in and they'll be deleted in the PR for #357?

PR #378, opened 2022-03-02
Feat 50 populate documents from cclw

opyate:
Also add documents alongside actions when importing a CCLW CSV.
- use `httpx` when creating an action, to make async checks against remote docs
- save all docs, but have a validity status against each
- validity statuses are fine grained, as some of those can potentially be retried at a later date
- it imports 2268 actions, which is what I had before this piece of work
- we're temporarily recreating the DB between each test, as rollback doesn't work (we commit manually in the service later, and this will be fixed)
- when the docker stack starts, it will start an import automatically


opyate:
Tests fail because of issue on Github's side:

image

loader/app/load/main.py
```diff
@@ -1,51 +1,91 @@
...
             "geography_id": geography_id,
             "type_id": action_type_id,
             "source_id": 1,  # CCLW is source_id 1
-            "documents": [],
+            "documents": [
+                {
+                    "name": doc.doc_name,
+                    "language_id": get_language_id_for_doc(doc),
+                    "source_url": doc.doc_url,
+                    "s3_url": None,  # TODO is this supposed to be doc.doc_url?
```
chrisaballard, line 50: @opyate the original idea for `s3_url` was that this stores the url of the document on s3 after it has been uploaded to an s3 bucket. It provides a link between the db entry for the action-document and the associated file on s3. `source_url` stores the original url of where the document was obtained from (e.g. CCLW or another source url)

chrisaballard:
@opyate just wanted to check my understanding form looking through the code is correct - currently the document entry is only created in the db and the document is not being uploaded to s3?

chrisaballard:
@ingemar @opyate I think something we'll need to consider in the future post-alpha is ensuring document urls posted to us are secure. i.e. it comes from a secure source and the document we receive is checked for malware before we store and process it.

loader/app/transform/util.py
```diff
@@ -39,49 +47,82 @@ def split_and_merge_urls(doc_urls, sep):
     return merged_doc_urls
 
 
-def get_urls(dataframe, policy_id=None, sep=';', sub_sep=None, url_sep=','):
-    country_code = dataframe['country_code']
+def get_policy_data(dataframe, sep=';', sub_sep=None, url_sep=',') -> Optional[Tuple[Key, PolicyData]]:
     policy_name = dataframe['policy_name']
+    country_code = dataframe['country_code']
+    policy_type = CCLWActionType[dataframe["policy_type"]].value
```
chrisaballard, line 53: I note that this mapping and the possible values for `Action Type` are hardcoded in code. If our definitions of Action Type change, we'll need to remember to update the mapping in `CCLWActionType`. Perhaps no obvious was round this, but just wanted to mention it.

loader/app/load/main.py
opyate: > f so, we'll need to make sure that we tackle this.

Sure, I'll start on #230 once this PR is merged.

opyate:
> If so, we'll need to make sure that we tackle this.

Sure, I'll start on #230 once this is merged.

loader/app/transform/util.py
opyate: Yes, this is just hard-coded for alpha. Ideally, it would be managed via DB or something.

opyate:
> @ingemar @opyate I think something we'll need to consider in the future post-alpha is ensuring document urls posted to us are secure. i.e. it comes from a secure source and the document we receive is checked for malware before we store and process it.

Yes: https://github.com/climatepolicyradar/navigator/pull/378/files#diff-1817be4c4b360eb646bf91d67c2b1b0b479073a6e45c385d981a9ee3335fead5R129

ingemar:
Hi! I think you tagged the wrong @ingemar, i.e me.

.env.example
```diff
@@ -21,3 +21,5 @@ SECRET_KEY=super_secret
 # machine user account for loader
 MACHINE_USER_LOADER_EMAIL=loader
 MACHINE_USER_LOADER_PASSWORD=password
+# machine user token has no expiry
+MACHINE_USER_LOADER_JWT=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJsb2FkZXIiLCJwZXJtaXNzaW9ucyI6ImFkbWluIn0.GRymxg-mbw2oO2zEBQm8p8eS7MqUSvBolljWgTufEcI
```
ingemar, line 25: Are you sure you want this key public to the world?

ingemar:
This PR is huge! Is it possible to break up in more atomic parts?

.env.example
opyate: Thanks for raising the concern! As this is just the example `.env` file, it should be OK.

opyate:
> This PR is huge! Is it possible to break up in more atomic parts?

This code base is in its early stages, and we're a tiny team, so we'll be delivering large chunks of work for the foreseeable in the race to alpha. But, otherwise - yes, that's solid advice, and we'll definitely make smaller changes in future! :)

chrisaballard:
When running `docker-compose up loader` the following error occurs:

```text
chrisballard@DESKTOP-B390QVR:~/code/cpr/navigator$ docker-compose up loader
navigator_postgres_1 is up-to-date
Recreating navigator_backend_1 ... done
Recreating navigator_loader_1  ... done
Attaching to navigator_loader_1
loader_1    | 2022-03-04 13:58:37,443 [WARNING] /app/app/main.py: Waiting 10 seconds for backend...
loader_1    | 2022-03-04 13:58:47,454 [WARNING] /app/app/main.py: Wait complete. Starting...
loader_1    | Traceback (most recent call last):
loader_1    |   File "/app/app/main.py", line 83, in <module>
loader_1    |     main()
loader_1    |   File "/app/app/main.py", line 63, in main
loader_1    |     load(policies)
loader_1    |   File "/app/app/load/main.py", line 17, in load
loader_1    |     geography_id = get_geography_id(country_code)
loader_1    |   File "/app/app/load/api_client.py", line 19, in get_geography_id
loader_1    |     return _get_id(country_code, get_geographies_lookup, "geography_id")
loader_1    |   File "/app/app/load/api_client.py", line 44, in _get_id
loader_1    |     lookup = lookup_fn()
loader_1    |   File "/app/app/load/api_client.py", line 17, in get_geographies_lookup
loader_1    |     return _get_lookup("geographies", "country_code")
loader_1    |   File "/app/app/load/api_client.py", line 69, in _get_lookup
loader_1    |     lookup[datum[lookup_key]] = datum
loader_1    | TypeError: string indices must be integers
navigator_loader_1 exited with code 1
```

PR #384, opened 2022-03-09
Feature/react query

eurolife:
Added react-query for state management. Not much difference to be seen in viewing the front end, but app state and state lifecycle is handled better. Geography, language, action type and source are now loaded using react-query. Action data is also submitted using react-query. 

Will still need some examples on re-using the auth token throughout the app without having to repeat the headers for every api request as per our discussion @ingemar-cpr


opyate:
> without having to repeat the headers for every api 

IIRC, you can set global config on axios:

```
axios.defaults.headers.common['header'] = 'value'
// or just for POST
axios.defaults.headers.post['header'] = 'value'
```

Or create a shared session:

```
let session = axios.create({
  headers: {
    common: {
      header: 'value'
    },
    post: { // or just for POST
      header: 'value'
    }
  }
})

// or after the session is instantiated
session.defaults.headers.post['header'] = 'value'

// or using Interceptors
session.interceptors.request.use(config => {
  config.headers.post['header'] = 'value';
  return config;
});
```

eurolife:
@opyate Thanks for that. The problem is, it also needs to get the auth token from local storage and when storing it in a separate instance as shown in your second example, it will error because localStorage or window.localStorage isn't defined because it's calling on it before the DOM loads.

opyate:
> failing test

It fails for me locally with `make test_frontend`.

eurolife:
When I'm inside the e2e folder, I run npx cypress run and the tests pass. What is different when running make test_frontend?  Is there something I need to change locally?

opyate:
> make test_frontend

CI runs `make test_frontend`, so if you make any changes, check that `make test_frontend` works too.

>  Is there something I need to change locally?

It depends on the changes you made. I'm trying to spot the issue...

opyate:
I checked out commit [2d3959b](https://github.com/climatepolicyradar/navigator/commit/2d3959b667a7f47fcd8bc757abc8b7dcd2f68e32) and `make test_frontend` passes (that's the last commit I made to get e2e to work with docker).

Will try looking at diffs.

opyate:
I'm seeing this when I run frontend locally with `make start_containers`:

image

opyate (approved):

PR #386, opened 2022-03-10
Fix mypy errors & enable checks for backend

joel-wright:
Enhance type checking to improve code quality:
- Fix existing type errors in backend code
- Enable mypy checking on backend code in pre-commit hook
- Update schema in places where the current types do not express the intent (e.g. optional types for non-optional fields)


backend/app/api/api_v1/routers/actions.py
```diff
@@ -65,18 +65,14 @@ async def action_create(
...
     )
 
     try:
         db_action = create_action(db, action_create)
     except Exception as e:
         if isinstance(e, IntegrityError):
-            raise HTTPException(
-                409, detail=f"Database integrity error, underlying={e.orig}"
-            )
+            raise HTTPException(409, detail=f"Database integrity error, underlying={e}")
```
opyate, line 75: Might be useful to know what [orig](https://docs.sqlalchemy.org/en/14/core/exceptions.html#sqlalchemy.exc.DBAPIError) is?

> The wrapped exception object is available in the [StatementError.orig](https://docs.sqlalchemy.org/en/14/core/exceptions.html#sqlalchemy.exc.StatementError.orig) attribute. Its type and properties are DB-API implementation specific.

backend/.pre-commit-config.yaml
```diff
@@ -23,5 +23,6 @@ repos:
     rev: v0.931
     hooks:
       - id: mypy
+        files: ^backend/
```
opyate, line 26: I guess we add `loader` and `pipeline` later?

backend/app/api/api_v1/routers/actions.py
opyate: Actually, this doesn't matter for now. We need to revise the error-handling at some point 👍🏻

backend/.pre-commit-config.yaml
joel-wright: yeah, we'll have to, they have a good chunk of errors themselves

backend/app/api/api_v1/routers/actions.py
joel-wright: I spent quite a while trying to get the type checker to accept that `.orig` but it stubbornly refused to believe it existed. The other option is to just disable the type check on that line...

opyate (approved):

PR #390, opened 2022-03-14
Fix mypy issues in pipeline

joel-wright:


pipeline/pdf2text.py
```diff
@@ -68,9 +68,14 @@ def get_pdf_files(
         s3_path_components = split_s3_path(
             pdf_path, include_bucket=True, include_filename=False
         )
-        for s3_document in s3_client.list_files(s3_path_components.bucket):
+        s3_documents = s3_client.list_files(s3_path_components.bucket)
+        if s3_documents is False:
+            raise RuntimeError("Failed to list s3 bucket '{s3_path_components.bucket}'")
```
kdutia, line 73: Should this be an f-string?

kdutia:
Changes look good to me, not tested locally though. From what I can tell there should be minimal conflict with @stefl14's work

joel-wright:
> Changes look good to me, not tested locally though. From what I can tell there should be minimal conflict with @stefl14's work

That's good news, I'll fix the f-string & have a chat with @stefl14 in the morning. Hopefully the test suite has caught any major issues.

Issue #391, opened 2022-03-14
Write text transformation function for pdf2text output to index


kdutia:
Also includes amending e.g. list outputs from the postprocessing tasks.

PR #392, opened 2022-03-15
Feat 355 group semantic lists

stefl14:
@kdutia

Pushing for visibility. Still a few TODOs sitting around that it'd be good to get some commentary on, plus I haven't merged in @joel-wright changes yet, waiting for it to be merged into dev.

@joel-wright you can see the nested json's fairly ugly (logic in _group_list_element method).

Class mainly a namespace atm (lots of static methods with not mutable state), but left it this way for stylistic consistency atm and possible extensibility with other features.


pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,268 @@
...
+        """
+        new_custom_attributes = {new_attribute: True}
+        if text_block["custom_attributes"] is not None:
+            text_block["custom_attributes"].update(new_custom_attributes)
+        else:
+            text_block["custom_attributes"] = new_custom_attributes
+        return text_block
+
+    @staticmethod
+    def _postprocess_list_grouped_page(text_blocks: List[dict]) -> dict:
```
kdutia, line 99: Maybe this function would be better renamed to something along the lines of `_remove_unmerged_lists`, as it seems to do only that?

kdutia, line 99: Maybe this function would be better renamed to something along the lines of `_remove_unmerged_lists`, as it seems to do only that?

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,268 @@
...
+                            #  validate the assumption that the previous block is context by pretty printing the list
+                            #  when coding.
+
+                            # Heuristic: If the list block is adjacent to the previous list block and the current
+                            # list id was unpopulated and the start of this conditional, assume we are on the same
+                            # list but on a new page, appending the appropriate metadata to custom attributes. We
+                            # likely need to add some logic to handle the case where the list is more than 1 element
+                            # away because of garbage text blocks at the end of the page. But for now this may be
+                            # good enough.
+                            # TODO: Perhaps add some more nuance here, as another possibility is that it's contiguous
```
kdutia, line 216: Not sure what you mean by this?

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,268 @@
...
+                                )
+
+                        dd[current_list_id].append(text_block)
+                        prev_list_ix += 1
+                    else:
+                        new_text_blocks.append(text_block)
+                else:
+                    new_text_blocks.append(text_block)
+                previous_block = text_block
+                blocks_seen += 1
```
joel-wright, line 235: it's not clear to me why this is being pre- and post-incremented in the loop

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,268 @@
...
+                    list_group = self._find_first_occurrence(
+                        self.regex_pattern, text_block["path"]
+                    )
+                    if list_group:
+                        current_list_id = f"{ix}_{list_group}"
+                        # Handle the case where we have a new list at the beginning of a page and where
+                        # the previous list block is assumed context.
+                        if (
+                            text_block["text_block_id"].split("_")[1] == "b1"
+                        ) and previous_block:
```
joel-wright, line 197: I don't understand how this test works; if the text block we're currently holding is the first element on a page, aren't we guaranteed that `previous_block` is `None`?

joel-wright:
I think I need to spend more time with the example data you send to properly understand what's going on here!

stefl14:
@kdutia This has become a rather convoluted branch at this point, which is mostly an artefact of my constantly discovering new interdependencies between features and bugs along the way. I've merged changes in hyphenation and style processing together
with the list processing stuff in this branch. 

To signpost things a little, at the moment reformat_adobe_output.py calls a text styling
post processor, then the hyphenation processor, and finally the list grouping processor. This ordering solved most of the obvious problems. I keep finding new issues and edge cases, but at this point it's likely best to stop developing as the code will
almost certainly become less comprehensible to anyone who isn't me, and this is likely to be an iterative testing process over
time anyway when we actually use it, so will stop engineering.

Code quality could certainly be a lot better, so have at it with comments if something isn't comprehensible or should be refactored.

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,634 @@
...
+
+        """
+        x_min = min(coord[0][0] for coord in coords)
+        y_min = min(coord[1][1] for coord in coords)
+        x_max = max(coord[1][0] for coord in coords)
+        y_max = max(coord[3][1] for coord in coords)
+        return [[x_min, y_min], [x_max, y_min], [x_min, y_max], [x_max, y_max]]
+
+
+class HyphenationPostProcessor:
```
stefl14, line 33: This class relies on english words library to decide how to deal with hyphenation. Not robust to
other languages. Any ideas on how to deal with this?

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,634 @@
...
+            if merged_block_text == []:
+                merged_block_text = text_block["text"]
+            else:
+                merged_block_text[-1] = merged_block_text[-1] + text_block["text"][0]
+                merged_block_text += text_block["text"][1:]
+
+            # Append style metadata that will later be added to custom attributes for text block.
+            if block_styling:
+                last_style_ix = len(merged_block_text[0].strip()) - 1
+                style_spans.append(
```
stefl14, line 182: Spacy styling here, as discussed. Turned out to be much simpler. Do we want to do something similar for lists? At the moment, the I've formatted lists in 2 ways - with html -esque formatting in the text block itself and with pretty printing in the metadata.

pipeline/pyproject.toml
```diff
@@ -10,6 +10,9 @@ PyPDF2 = "^1.26.0"
 pdfservices-sdk = "^1.0.1"
 navigator = {path = "../common", develop = true }
 tqdm = "^4.62.3"
+numpy = "^1.22.3"
+pandas = "^1.4.1"
+english-words = "^1.1.0"
```
stefl14, line 15: only handles english.

pipeline/processor/postprocessor.py
kdutia: Maybe using a dataset accessible from `nltk.corpus` could be better, e.g. [wordnet](https://www.nltk.org/howto/wordnet.html)? NLTK also contains corpora in other languages we could extend to later, but the language issue isn't too big a deal for now as Adobe Extract is only English language.

pipeline/processor/postprocessor.py
kdutia: I think lists make sense as one line (string in `TextBlock.text` list) per bullet - how would you imagine using this styling approach for lists?

Lists were structured pretty well last time I saw the code (haven't checked this time though). As far as I can remember the only potentially outstanding issue was nested lists.

pipeline/pyproject.toml
kdutia: I suggest you use nltk as above - you'll likely need to add [`nltk.download("wordnet")`](https://stackoverflow.com/questions/68718901/what-does-nltk-downloadwordnet-accomplish) or similar to the Dockerfile too.

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,634 @@
...
+
+import pandas as pd
+from english_words import english_words_set
+
+from pipeline.extract.document import Document, TextBlock
+
+
+class PostProcessor:
+    @staticmethod
+    def _minimal_bounding_box(coords: List[list]) -> list:
```
kdutia, line 15: Typing could be more specific here

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,634 @@
...
+                    li[ix] = word_fragment + l
+                # Reset.
+                current = None
+            if regex_match:
+                # Strip matching regex from the end of the string.
+                li[ix] = l[: regex_match.start()]
+                current = regex_match[0]
+        return li
+
+    def process(self, contents: dict) -> dict:
```
kdutia, line 80: For the `process` methods, I'm thinking that even if the underlying logic operates on dicts, they should still take and receive `Document` objects? This way, if we add field validation to the `Document` class it's automatically built into the postprocessors.

It should be pretty easy to define a `Document.to_dict()` method based on `Document.save_json()`. I'm not sure about the effort involved in a `Document.from_dict()` classmethod though, or how much effort it would take to input and output `Document` objects through your code. Happy for you to make the call on this.

kdutia:
@stefl14 Looks good - I think some tests on each postprocessor will help explain to others/us in future what each of them does.

On your points:
* I wouldn't worry too much about non-english for now but I've suggested you use wordnet and NLTK anyway, as it's a widely used dictionary and NLTK supports other languages. Our language skills might prove the biggest challenge when we start to try to identify hyphenated words in other languages ;)
* As mentioned in an inline comment I think it'd be really useful if each processor inputs and outputs a `Document` object, which should be pretty easily doable using `Document().to_dict()` and `Document.from_dict()` methods. I'm not sure how simple the latter is though, so don't spend too much time on it if it proves a hassle.

pipeline/processor/postprocessor.py
stefl14: @kdutia  Yes so I refactored it a bit. Nested lists are handled now with html-esque tags (so in my formatting there is
li1, li2, etc at the moment, and nesting is dealt with via tabulation and detabulation in the pprint function. Should be relatively easy to refactor the html-esque styling into something else with what I've got now if need be.

As far as the text goes, there is one line per bullet as you suggest, but I have something like <Lbl>(a)<\Lbl><Lbody>...<\Lbody>. ofc I could have normalised the labels but thought in some cases knowing an (a) from a bullet is useful, though though metadata on nesting level is probably enough.

There are still some issues the logic occasionally hits like deciding whether to interpret a span as a bullet or italics in a list body, as both cases occur. But since I kept finding issues like this that crop up occasionally, I've ignored them as I don't have true visibility on how common they are (quite hard to tell without manual work).

kdutia: Let's not worry about edge cases for now. We'll have a better grip on how prevalent they are when we run these postprocessing steps on the whole corpus, and we can do that in #391.

Maybe it's worth creating a github markdown or Notion page giving examples of inputs and outputs for each of the steps, with some of the details you've mentioned here? I get the gist of your comments but am finding it hard to visualise them without examples, which probably means we'll find these comments hard to decode in a couple of months too.

stefl14: Yep definitely agree, will set up a notion page.

pipeline/processor/postprocessor.py
stefl14: Easy fix, added a bunch of class methods to handle.

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,634 @@
...
+from copy import deepcopy
+from typing import List, Dict, Optional
+
+import pandas as pd
+from english_words import english_words_set
+
+from pipeline.extract.document import Document, TextBlock
+
+
+class PostProcessor:
```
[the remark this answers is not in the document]
stefl14: Done.

pipeline/extract/document.py
```diff
@@ -123,7 +127,30 @@ def save_text(self, text_filepath: Path):
         with open(text_filepath, "wt") as f:
             f.write(self.to_string())
 
+    @classmethod
+    def from_dict(cls, json_dict: dict) -> "Document":
+        """Load a document from a dictionary"""
+        for page in json_dict["pages"]:
```
stefl14, line 133: Fixed bug where the text blocks within pages were not conforming ot the Doc>Page>TB format.

pipeline/extract/document.py
```diff
@@ -123,7 +127,30 @@ def save_text(self, text_filepath: Path):
...
+            page["text_blocks"] = [TextBlock(**block) for block in page["text_blocks"]]
+        return cls(
+            pages=[Page(**page) for page in json_dict["pages"]],
+            filename=json_dict["filename"],
+        )
+
     @classmethod
```
stefl14, line 140: Implement method to convert from json to document.

pipeline/reformat_adobe_output.py
```diff
@@ -0,0 +1,47 @@
+import argparse
```
stefl14, line 1: Just a simple CLI for running locally. Can refactor and add s3 option if needed but didn't bother for now as it's super quick and can be ran locally for now.

pipeline/test/test_processor.py
```diff
@@ -0,0 +1,790 @@
...
+def adobe_text_styling_postprocessor() -> AdobeTextStylingPostProcessor:
+    return AdobeTextStylingPostProcessor()
+
+
+@pytest.fixture
+def adobe_hyphenation_postprocessor() -> HyphenationPostProcessor:
+    return HyphenationPostProcessor()
+
+
+def test_hyphenation_postprocessor_rewrap_hyphenated_words(document):
```
stefl14, line 39: I deliberately chose an example that this processor doesn't handle perfectly to illustrate some of the arbitrariness of the Adobe results (see comments).

pipeline/test/test_processor.py
```diff
@@ -0,0 +1,788 @@
...
+def adobe_text_styling_postprocessor() -> AdobeTextStylingPostProcessor:
+    return AdobeTextStylingPostProcessor()
+
+
+@pytest.fixture
+def adobe_hyphenation_postprocessor() -> HyphenationPostProcessor:
+    return HyphenationPostProcessor()
+
+
+def test_hyphenation_postprocessor_rewrap_hyphenated_words(adobe_hyphenation_postprocessor):
```
stefl14, line 39: I deliberately used a case here that has some arbitrary problems (truncating at hyphens for some blocks, not others). The functionality I'm testing still passes, but I thought it would be useful to show just how arbitrary the Adobe API's outputs can be. This can also be seen in some other test cases. See comments for more detail.

pipeline/test/test_processor.py
```diff
@@ -0,0 +1,788 @@
...
+    expected_output_2 = "L"
+    actual_output_2 = adobe_list_postprocessor._find_first_occurrence(regex, input_2)
+    assert actual_output_2 == expected_output_2
+    # Case 3: No matching regex.
+    input_3 = ["Document", "P"]
+    actual_output_3 = adobe_list_postprocessor._find_first_occurrence(regex, input_3)
+    assert not actual_output_3
+
+
+def test_adobe_text_styling_merge_test_blocks(adobe_text_styling_postprocessor):
```
stefl14, line 101: If you visit the actual document from the test case, it shows that the logic isn't particularly useful in this particular case. As above, I chose this case to further illustrate the combinatorial possibility of cases to deal with in the Adobe API. Nevertheless, tested functionality passes.

pipeline/test/test_processor.py
```diff
@@ -0,0 +1,788 @@
...
+        "<Lbl>(2)<\\Lbl><LBody>The failure of a licensee to promptly inform the P.P.U.C. in writing of any technical changes to the renewable energy system that affects any of the above may, at the P.P.U.C. discretion, result in a fine of not more than two hundred dollars ($200).<\\LBody>\n\n<\\li2>\n",
+        "<Lbl>(c)<\\Lbl><LBody>shall, at its own-expense, make available to each of its eligible customer generators who have installed a net metering system the meter (or set of meters) that is needed to determine the net flow of electricity both into and out of the electricity grid;<\\LBody>\n\n<\\li1>\n",
+    ]
+
+    actual_out = adobe_list_postprocessor._format_semantic_lists(in_df)
+    assert actual_out == expected_out
+
+
+def test_adobe_list_processor_pprinter(adobe_list_postprocessor):
```
stefl14, line 396: This test is more so you can just see what the pretty printed output would look like (if used by, say, the API) when running the test suite with the arguments -rP

pipeline/extract/document.py
kdutia: Could this use `from_dict` after `data = json.load(f)`?

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,663 @@
...
+                    new_string += f"<LBody>{text}<\LBody>\n" + f"\n<\li{list_number}>\n"
+                else:
+                    new_string += f"<LBody>{text}<\LBody>\n"
+                lst.append(new_string)
+        return lst
+
+    @staticmethod
+    def _duped_line_reformatting(line: str) -> str:
+        """
+        Helper to remove duplication in pprinter method.
```
kdutia, line 383: It'd be good to be a bit more specific here about 'removing duplication' means

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,663 @@
...
+    def process(self, doc: Document, filename: str) -> Document:
+        """
+        Parse the elements belonging to a list into a single block, including the list's introductory text.
+
+        Args:
+            filename:
+            doc: dict.
+
+        Returns:
+            A dictionary with values corresponding to a semantic block of a list (introductory context plus the list itself).
```
kdutia, line 657: I think this docstring needs updating

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,663 @@
...
+            return "subscript"
+        elif text_block["custom_attributes"].get("TextDecorationType") == "Underline":
+            return "underline"
+        elif text_block["custom_attributes"].get("TextPosition") == "Sup":
+            return "superscript"
+        else:
+            return None
+
+    @staticmethod
+    def _add_text_styling_markers(text: str, styling: str) -> str:
```
kdutia, line 143: It looks like this method can be deleted as we're using the spacy-like styling objects now?

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,663 @@
...
+        return contents
+
+
+class AdobeTextStylingPostProcessor(PostProcessor):
+    """
+    Some semantic passages have separate blocks for styling markers such
+    as underlines, superscripts and subscripts. We want to group such cases
+    into single contiguous blocks. However, we want to keep the styling information
+    because it's often relevant to semantics. For example, CO2 is often represented
+    using a subscript. For now, we keep everything and indicate the styling with
```
kdutia, line 111: I think the 'indicating the styling with HTML tags inline' part of this docstring needs updating

pipeline/test/test_processor.py
kdutia: 👍🏼

pipeline/processor/postprocessor.py
```diff
@@ -0,0 +1,663 @@
...
+    """Further processing of processed outputs from the Adobe API (handling cases not easily
+    handled by the API itself)."""
+
+    regex_pattern = r"L\[?\d?\]?$"
+
+    def __init__(self):
+        super(AdobeListGroupingPostProcessor, self).__init__()
+
+    @staticmethod
+    def _find_first_occurrence(regex_pattern: str, list_of_strings):
```
kdutia, line 285: As this method and the method below are always used to find the *list* occurrences, and are almost identical in their implementation, I'd suggest the following changes:
• rename `self.regex_pattern` to `self.list_regex_pattern`
• `_find_all_occurrences` -> `_find_all_list_occurrences`
• delete `_find_first_occurrence` and instead use `self._find_all_list_occurrences(list_of_strings)[0]` in your code. Alternatively rename `_find_first_list_occurrence` and set its implementation to use `_find_all_list_occurrences`

pipeline/test/test_processor.py
kdutia: this is great!

pipeline/test/test_processor.py
```diff
@@ -0,0 +1,788 @@
...
+                {"style": "underline", "start_idx": 21, "end_idx": 29},
+            ]
+        },
+    }
+
+    actual_output = adobe_text_styling_postprocessor.merge_text_blocks(example)
+    assert actual_output == expected_output
+
+
+def test_adobe_list_processor_semantic_lists(adobe_list_postprocessor):
```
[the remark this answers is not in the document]
kdutia: As discussed we might want to change the way we mark up list labels, list elements and lists in future, but the important thing for now is that we've got code that gives us these. This looks great.

pipeline/pyproject.toml
kdutia: Makes sense- Adobe can only handle English so we don't need to worry about building post-processors for other languages.

pipeline/reformat_adobe_output.py
kdutia: 👍🏼 I imagine we'll look at how we separate out the Adobe API call, `data_to_document` step and this with @joel-wright at a later date.

kdutia:
@stefl14 LGTM after comments and the remainder of the tests are resolved. Probably also worth checking it runs on the full corpus without failing if you haven't, before we merge it.

PR #395, opened 2022-03-16
Feature/test actions api

eurolife:
This branch starts to set up user authentication with React Query and the react-query-auth library. It also uses React Query to load and store actions from the actions API. View actions at http://localhost:3000/actions (this will load a list of 20 actions from the api). This is not a final page and was mainly done just to test the api.

Further user authentication and account screens and methods will be done in a separate dedicated branch.

@opyate please note that there are some html tags returned in the descriptions that I think need to be stripped out?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,7 @@ dist/
 downloads/
 eggs/
 .eggs/
-#lib/
+lib/
 lib64/
 parts/
 sdist/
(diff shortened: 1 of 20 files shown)


opyate:
> @opyate please note that there are some html tags returned in the descriptions that I think need to be stripped out?

OK, I'll tackle that next: #396

PR #401, opened 2022-03-17
Feat 351 get action

opyate:
Get action by ID.

### DIFF
--- a/backend/app/api/api_v1/routers/actions.py
+++ b/backend/app/api/api_v1/routers/actions.py
@@ -4,21 +4,42 @@ from sqlite3 import IntegrityError
 from typing import Optional
 
 import httpx
+from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi_pagination import Page
+from fastapi_pagination.ext.sqlalchemy import paginate
+
 from app.core.auth import get_current_active_user
-from app.db.crud.action import create_action, get_actions_query, is_action_exists
+from app.db.crud.action import (
+    create_action,
+    get_actions_query,
+    is_action_exists,
+    get_action,
+)
 from app.db.crud.document import create_document
 from app.db.models.document import DocumentInvalidReason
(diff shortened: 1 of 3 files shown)


joel-wright (commented):
It would be nice if we also had tests validating the behaviour when the action does not exist, when the request is supplied with invalid auth headers & no auth headers.

opyate:
> It would be nice if we also had tests validating the behaviour when the action does not exist, when the request is supplied with invalid auth headers & no auth headers.

Indeed, but I think auth is tested once somewhere, and 404 is just a thing you perhaps have to trust that starlette/fastapi codebases have already tested?

joel-wright:
> > It would be nice if we also had tests validating the behaviour when the action does not exist, when the request is supplied with invalid auth headers & no auth headers.
> 
> Indeed, but I think auth is tested once somewhere, and 404 is just a thing you perhaps have to trust that starlette/fastapi codebases have already tested?

I'm not sure I follow your reasoning here - all the API calls must be authenticated and it looks like we're enforcing that with an endpoint specific `Depends()`, so we need to verify that all the API endpoints require auth. It's not a case of testing whether the auth API works, but that these specific API calls have the appropriate authentication requirements enabled. Have I missed a test elsewhere that validates that all our endpoints require appropriate auth?

Also, it doesn't look to me like starlette/fastapi are handing the 404 - we are explicitly implementing the 404 response in our own `get_action()` function and there's no test for that code path, so how do I know the code is working as expected?

opyate:
> we are explicitly implementing the 404 response in our own `get_action()` function

Ah, sorry - I read that as testing missing endpoints, or verifying that endpoints exist.

> all the API calls must be authenticated and it looks like we're enforcing that with an endpoint specific Depends()

Noted. Test coming up!

joel-wright (approved):
LGTM

PR #405, opened 2022-03-17
decouple loader from stack. Fixes #404

### LINKED ISSUE #404: Decouple loader from stack initialisation
Currently, we have a catch 22 where the loader starts up before migrations run (after deleting volumes), and loader fails by not finding geographies.

Loader will be moved to a separate task, and not run immediately as part of stack startup.

opyate:


eurolife (approved):

joel-wright (approved):
LGTM

Issue #348, opened 2022-02-14
Use cli to run text extraction for CCLW corpus using Adobe Extract API and Embedded Text parser


chrisaballard:
Use the pdf2text cli to extract the text from the english language PDFs in the CCLW corpus and generate the json output files in an s3 bucket. This will create a dataset which will allow us to progress with search and ETL of text data.

- [x] make a list of PDFs from english language CCLW, based on analysis from prototype
- [x] run these through the extract pipeline
- [x] when postprocessing done, rerun `data_to_document` step on all intermediate files

chrisaballard:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Product-and-Tech-Team-613867967a1a9200123c9aad/issues/climatepolicyradar/navigator/348?planning-poker) with ZenHub @eurolife @kdutia @opyate

kdutia:
@ingemar-cpr @stefl14 and I talked about doing a one-off data cleaning exercise for CCLW to get the document title from each PDF. Before doing this data load we should check that the data is suitable for this.

kdutia:
I emailed adobe today regarding outstanding BAD_PDF and 500 errors, and the timeout and variable page splitting limit issues we were having.

PR #411, opened 2022-03-21
Feat 348 - pdf2text improvements

kdutia:
This PR contains tweaks and improvements to the pdf2text CLI made when running the whole English language CCLW PDF corpus through it, as part of #348. Namely:

1. The max page limits are decreased as these were causing PDFs with many tables to be rejected by the API, even after a split
2. Addition of v. high read and connection timeouts (essentially `=None` in requests, but the Adobe SDK doesn't allow this value), to avoid ReadTimeoutErrors and WriteTimeoutErrors
3. Rotation of text elements as part of `AdobeAPIExtractor` is now measured relative to the page's rotation (e.g. 90º text rotation on a page that is 90º rotated = 0º rotation). I've also raised bug issue #413 as fixing this issue highlighted that line detection is broken for rotated PDFs.
4. Multiprocessing has been added to the CLI to speed up the otherwise slow process. As discussed with @joel-wright we'll  likely have a different multiprocessing implementation in the deployed data pipeline.

There are also a bunch of docstring changes because I branched the version of dev with restrictive docstring formatter rules 😁

**Update:** I've also added two notebooks. One is for us to keep a log of the Adobe errors whilst we resolve it with the Adobe team, and the other is to be able to run the `data_to_document` step on a set of Adobe and pdfalto outputs to produce a new set of json and txt files, as we work on the pipeline and postprocessing steps. The second notebook is required at the moment mainly as the `intermediate` directory contains multiple folders or files per PDF, and I haven't turned it into another CLI because I imagine we'll end up refactoring the CLI to separate the `pdf_to_data` and `data_to_document` processes as we look to deploy it in a workflow management system.


pipeline/Dockerfile
```diff
@@ -6,8 +6,6 @@ WORKDIR /app
 RUN apt update && \
     apt install -y cmake
 
-RUN mkdir ~/.ssh
-RUN ssh-keyscan -H github.com >> ~/.ssh/known_hosts
```
kdutia, line 10: These ssh commands were mistakenly left in the Dockerfile from another PR when we tried to perform `git submodule update` for pdfalto through SSH. We switched to https as on line 13/11 as it was easier in CI.

opyate (approved):

Issue #414, opened 2022-03-23
Run pdf2text CLI on documents loaded since prototype


kdutia:
This also includes:
* appropriately renaming the Adobe outputs created using prototype filenames 
* running language detection on all new documents
* removing the dependency of the search indexer on the processed_policies CSV file (see PR #419)

PR #419, opened 2022-03-28
Feat 393 - opensearch search and indexing example scripts

kdutia:
This PR implements the following:
* a docker-compose setup for the indexing scripts, one Opensearch node, and Opensearch dashboards
* a `text2embeddings.py` CLI to load JSON data produced by the `pdf2text` pipeline and convert it to embeddings (.memmap) and text and IDs (.json) files.
* an `index_data.py` CLI to load document metadata from the Navigator database and the prototype *processed_policies.csv* file and index this data alongside the text and embeddings created using `text2embeddings` into the search index. At the moment the CSV file is used as the files processed through the pdf2text CLI used the prototype filenames. There will be a bit of manual work to process the files in the navigator repo and map prototype filenames to navigator filenames when we have PDFs stored on s3, which I've raised in #414.
* an `opensearch-query-example.ipynb` notebook that demonstrates running a query on the index. This is to be developed further and integrated into the Navigator APIs.

Gaps between this PR and product-level implementation as well as the various TODOs in the comments are:
* only one opensearch node is defined
* [non-prod docker-compose settings](https://opensearch.org/docs/latest/opensearch/install/docker/#sample-docker-compose-file-for-development) are used for opensearch
* KNN index settings need to be tuned (although these may be ok for first release, as we used them successfully in the prototype)

There also may be some devops considerations here as Opensearch won't run as smoothly as the database or likely in Github actions. There are no tests in this PR.

@stefl14 tagging you for a heads up but don't feel like you need to give it a detailed look


search-index/index_data.py
```diff
@@ -0,0 +1,248 @@
...
+    query = """
+        SELECT document_id, source_url, s3_url, document.language_id as document_language_id, document.name AS document_name, action.*, language.language_id, language.language_code, language.name as language_name, \
+        geography.*, source.source_id, source.name as action_source_name, action_type.action_type_id, action_type.type_name as action_type_name
+        FROM document
+        INNER JOIN action ON (document.action_id = action.action_id)
+        LEFT JOIN language on (document.language_id = language.language_id)
+        LEFT JOIN geography on (action.geography_id = geography.geography_id)
+        LEFT JOIN source on (action.action_source_id = source.source_id)
+        LEFT JOIN action_type on (action.action_type_id = action_type.action_type_id);
+        """
```
kdutia, line 35: Please can someone double-check this SQL query as it's one of the first I've ever written!

opyate: If it works for now, then great. This will change anyway, as we'll put data from external sources in a "stage 0" DB first, where the indexer will be reading from, and then data for public consumption will go to the next stage.

opyate (approved):

search-index/index_data.py
```diff
@@ -0,0 +1,248 @@
...
+def ensure_safe(url: str) -> str:
+    """Ensure a URL is safe.
+
+    Some documents use http, not https. Instead of just ignoring those,
+    we'll try download a doc securely, if possible.
+
+    # TODO: this function is also used in the loader. We use it here to repeat that transformation for a successful join. Do we care that it's copy & pasted?
+    """
+    if "https://" not in url:
+        url = url.replace("http://", "https://")
```
joel-wright, line 49: nit: Might be safer to do the following:

```suggestion
    if url.startswith("http://"):
        url = url.replace("http://", "https://")
    if not url.startswith("https://"):
        raise RuntimeException(f"Unsupported URL: {url}")
```

joel-wright (approved):
Lots of changes coming to the way Documents are stored in the navigator db, and how we handle intermediate processing state. This is all a great basis for getting information into the index & demonstrating how queries can be performed.

search-index/index_data.py
kdutia: Leaving this as-is as it's safest to do what the loader is doing for now.

Issue #393, opened 2022-03-16
Create Opensearch search and indexing example scripts


ingemar-cpr:
- basic dockerised opensearch instance
- script to index data out of pdf2text cli
- script to create API for search which matches alpha requirements, producing expected result
- also include e.g. index warmup

**Implementation**

- indexing script with its own docker environment which loads from the `document` table and relevant joined tables
- `/search` API endpoint 
- API interfaces (request, response) will be fully developed in a separate issue

**To do**
- [x] write SQL query to get correct data to index
- [x] CLI to create embeddings
- [x] indexer based on data and embeddings
- [x] design correct opensearch query - consider thresholding KNN using `function_score` here
- [ ] build a `/search` API around the query

PR #426, opened 2022-03-31
Add keyword and year range filters to Opensearch query example

kdutia:
Added keyword and year range filters to Opensearch query example notebook. Part of already closed issue #393.

@stefl14 FYI


joel-wright (approved):
Have taken a look at the PR & the additions are exactly what I needed, thanks :)

PR #427, opened 2022-03-31
Feat 415 auth endpoints

opyate:


backend/pyproject.toml
```diff
@@ -23,6 +23,8 @@ SQLAlchemy = "^1.4.31"
 SQLAlchemy-Utils = "^0.38.2"
 fastapi-health = "^0.4.0"
 uvicorn = {extras = ["standard"], version = "^0.17.6"}
+slowapi = "^0.1.5"
```
opyate, line 26: For rate limiting.

backend/pyproject.toml
```diff
@@ -23,6 +23,8 @@ SQLAlchemy = "^1.4.31"
...
+boto = "^2.49.0"
```
opyate, line 27: For checking if the app is currently running on AWS Cloud.

backend/app/api/api_v1/routers/tests/test_users.py
```diff
@@ -1,108 +1,65 @@
...
     response = client.put(
-        f"/api/v1/users/{test_superuser.id}",
-        json=new_user,
-        headers=superuser_token_headers,
+        "/api/v1/users/me",
+        json={
+            "email": test_user.email,
+            "names": "New name",
+            "is_active": test_user.is_active,  # TODO prevent user from setting these flags
+            "is_superuser": test_user.is_superuser,  # TODO prevent user from setting these flags
```
joel-wright, line 26: These seem pretty critical, can we open a bug ticket to make sure we don't lose this info!

backend/app/api/api_v1/routers/users.py
```diff
@@ -1,85 +1,52 @@
...
-    return edit_user(db, user_id, user)
 
+    updated_user = edit_user(db, current_user.id, user)
+    send_email(EmailType.account_changed, updated_user)
+    return updated_user
 
-@r.delete("/users/{user_id}", response_model=User, response_model_exclude_none=True)
-async def user_delete(
+
+@r.delete("/passwords/me", response_model=bool, response_model_exclude_none=True)
```
joel-wright, line 41: I have the same objection to using `DELETE` here as for the admin endpoint.

backend/app/api/api_v1/routers/users.py
```diff
@@ -1,85 +1,52 @@
...
-    current_user=Depends(get_current_active_superuser),
+    current_user=Depends(get_current_active_user),
 ):
-    """Deletes existing user"""
-    return delete_user(db, user_id)
+    """Deletes a password, and kicks off password-reset flow."""
+
+    deactivate_user(db, current_user.id)
+    activation_token = create_password_reset_token(db, current_user.id)
+    send_email(EmailType.password_reset_requested, current_user, activation_token)
```
joel-wright, line 51: There seem to be a couple of issues here:
- It looks like a user must be logged in in order to request that their password is reset because they've forgotten it, which seems like it could be an issue.
- The password reset form should be unauthenticated, take an email & if valid kick off a password reset flow for a user, this shouldn't immediately disable a user or it's open for abuse.

backend/app/core/security.py
```diff
@@ -32,3 +32,7 @@ def create_access_token(*, data: dict, expires_delta: Optional[timedelta] = None
     to_encode.update({"exp": expire})
     encoded_jwt = jwt.encode(to_encode, SECRET_KEY, algorithm=ALGORITHM)
     return encoded_jwt
+
+
+def get_password_reset_token_expiry_ts() -> datetime:
+    return datetime.utcnow() + timedelta(weeks=1)
```
joel-wright, line 38: Given that we're sending this by email, which isn't exactly known for security, a reset token valid for a week seems like a very long time.

backend/app/db/crud/user.py
```diff
@@ -20,56 +23,104 @@ def get_user_by_email(db: Session, email: str) -> t.Optional[User]:
...
+
+def create_password_reset_token(
+    db: Session,
+    user_id: int,
+) -> PasswordResetToken:
+    # TODO make configurable
+    future_date = get_password_reset_token_expiry_ts()
+
+    row = PasswordResetToken(
+        user_id=user_id, token=random_string(), expiry_ts=future_date
```
joel-wright, line 119: Related to the above: this 12 character random string seems a little arbitrary, why is this not a generated JWT activation token from which we extract the appropriate data?

backend/app/db/crud/user.py
```diff
@@ -20,56 +23,104 @@ def get_user_by_email(db: Session, email: str) -> t.Optional[User]:
...
     db.add(db_user)
     db.commit()
     db.refresh(db_user)
     return db_user
+
+
+def activate_user(
+    db: Session,
+    user: User,
+    activation_token: PasswordResetToken,
```
joel-wright, line 71: I feel like we should really be validating a generated secure JWT activation token here...

backend/app/main.py
```diff
@@ -43,16 +50,27 @@ async def db_session_middleware(request: Request, call_next):
 
 @app.get("/api/v1", include_in_schema=False)
 async def root():
-    return {"message": "Hello World"}
+    return {"message": "CPR API"}
```
joel-wright, line 53:
```suggestion
    return {"message": "CPR API v1"}
```

joel-wright (commented):
Some comments, a few things that feel like they need to be fixed.

backend/app/db/models.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+class User(Base):  # noqa: D101
+    __tablename__ = "user"
+
+    id = sa.Column(sa.Integer, primary_key=True)
+    email = sa.Column(sa.String, unique=True, index=True, nullable=False)
+    names = sa.Column(sa.String)
+    hashed_password = sa.Column(sa.String)
+    is_active = sa.Column(sa.Boolean, default=True, nullable=False)
```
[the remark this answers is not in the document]
opyate: Good idea, 4b6f39d

backend/app/db/schemas/user.py
```diff
@@ -7,26 +7,25 @@ class UserBase(BaseModel):  # noqa: D101
     email: str
     is_active: bool = True
```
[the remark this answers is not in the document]
opyate: Thanks, 292b8c8

backend/app/api/api_v1/routers/tests/test_admin.py
```diff
@@ -0,0 +1,184 @@
...
+    assert response.status_code == 401
+    response = client.put("/api/v1/admin/users/123")
+    assert response.status_code == 401
+    response = client.delete("/api/v1/admin/users/123")
+    assert response.status_code == 401
+
+
+def test_unauthorized_routes(client, user_token_headers):
+    response = client.get("/api/v1/admin/users", headers=user_token_headers)
+    assert response.status_code == 403
```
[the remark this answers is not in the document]
opyate: I agree, and have used 404 everywhere before too, probably for the same reasons as you (attacker knowing they've hit the right resource, but now just have to engineer access). Counter arguments are that those resources could be rate-limited, so dictionary attacks take forever, but I guess there are other ways to get the password (e.g. social engineering a known admin).
Happy to change it to 404.

backend/app/api/api_v1/routers/tests/test_users.py
opyate: #428

backend/app/api/api_v1/routers/users.py
opyate: True 👍🏻 Will modify.

backend/app/db/crud/user.py
opyate: Because it doesn't need to be? Just has to be unique, hard to guess.

Do you mean that using JWT means we can do away with the password_reset_token table? Store the expiry as data in the JWT? If so, then good idea.

backend/app/core/security.py
opyate: Oops, it's missing the "make configurable" TODO. I might as well just add it in now.

backend/app/api/api_v1/routers/tests/test_admin.py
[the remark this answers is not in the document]
joel-wright: I think I'd be happier with 404, we can always change it later if we decide to implement other mitigations.

backend/app/db/crud/user.py
opyate: But it might also be a good idea to keep track of these resets, especially tokens that expired and haven't been redeemed, so we can reach out to these users, and perhaps check if emails are bouncing, etc.

joel-wright: What I initially thought we could do was encode the user id & expiry in the token, then validate & extract them from it... I had thought we could do away with the table, but now that I think about it more that probably means we'd lose the ability to track when a token had been used (and invalidation on multiple requests etc) :disappointed: .

backend/app/api/api_v1/routers/tests/test_admin.py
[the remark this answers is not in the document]
opyate: Will action in #429

backend/app/api/api_v1/routers/admin.py
```diff
@@ -0,0 +1,110 @@
...
+    user_id: int,
+    db=Depends(get_db),
+    current_user=Depends(get_current_active_superuser),
+):
+    """Deletes a password for a user, and kicks off password-reset flow."""
+
+    deactivated_user = deactivate_user(db, user_id)
+    activation_token = create_password_reset_token(db, user_id)
+    send_email(EmailType.password_reset_requested, deactivated_user, activation_token)
+    return True
```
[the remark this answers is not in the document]
opyate: As per the discussion this morning, this is now RPC `POST /password-reset`

joel-wright (approved):
LGTM, just wanted to flag the styling change in the frontend code

PR #430, opened 2022-04-01
Feat 428 prevent users from setting flags

opyate:


backend/app/api/api_v1/routers/tests/test_users.py
```diff
@@ -21,8 +21,10 @@ def test_edit_user(mock_send_email, client, test_user, user_token_headers, test_
         json={
             "email": test_user.email,
             "names": "New name",
-            "is_active": test_user.is_active,  # TODO prevent user from setting these flags
-            "is_superuser": test_user.is_superuser,  # TODO prevent user from setting these flags
+            "is_active": not test_user.is_active,  # will be ignored
+            "is_superuser": not test_user.is_superuser,  # will be ignored
+            "affiliation_organisation": "org",
+            "affiliation_type": ["type 1", "type 2"],
```
joel-wright, line 27: Are these fields that a user should be able to set for themselves? Would we ever want to limit features by organisation?

joel-wright (approved):
Looks good, just the question about users setting their own affiliations (but for the alpha that probably shouldn't hold up this PR)

backend/app/api/api_v1/routers/tests/test_users.py
opyate: I think the drop-down on this page will become multi-select: https://climatepolicyradar.org/request-access

PR #435, opened 2022-04-05
Feat 391 text transformation function

stefl14:
I experimented with a bunch of more complex stuff but settled on relatively simple solutions in the end because they generalise
comparatively well.

This is best seen in how I handle lists, with very simple processing: any block subsequent to a list, irrespective
of its page, is considered context or part of the same list on the previous page. There isn't too
many failure modes so far as I can tell. I tried some more complex stuff but it deals with rare cases and creates other
errors, so left in TODOs for now.

You'll notice that some of the fixes are not strictly within scope of the task. The reason is that this task clarified some
of the upstream errors in postprocessing, which I decided to keep on this branch as they're relatively trivial. (new metadata
for debugging etc).

get_text_from_merged_block is where most of the interesting stuff happens, and this part is what led to most of the upstream fixes.

4) **One thing I noticed is that, as is, blocks are sometimes merged by the styling processor because they have the same path,
but there is an block between them with a different path but no stylespan attribute. This Adobe weirdness (italicisation is not considered a style, but a Span, for example) means that we end up missing italicised text from merged text blocks. As far as I can tell this is relatively rare, so I've not fixed it yet (left as TODO), but it's worth us assessing this in more detail because italics to me imply semantic importance. See comment in exception handling for CCLW example.**

I also noticed that we probably shouldn't be removing superscripts in the case where we have 1st,
2nd, etc as these are likely capturable semantics. Dealt with using a simple regex.

Unit testing not extensive as the functions are simple. But I tested the most complex function for clarity and to make it much easier to follow. There are 4 edge cases considered. Worth looking at.

### DIFF
--- a/search-index/test/test_text2embeddings.py
+++ b/search-index/test/test_text2embeddings.py
@@ -81,3 +81,77 @@ def test_get_text_from_merged_block():
 
     result_4 = text2embeddings.get_text_from_merged_block(input_4)
     assert result_4 == expected_4
+
+
+def test_get_text_from_list():
+    # Case 1: A nice complex nested list example with context.
+    previous_context_block = """The P.P.U.C.-
+* shall develop a standard contract providing for net energy metering, and shall, upon request, make this contract available to eligible customer-generators;
+* shall prepare appropriate technical standards for grid connection of renewable energy systems, and inspect and provide a license for those renewable energy installations that meet the technical standards developed by P.P.U.C. and the other provisions of this legislation. Issuance of a license shall be solely to show that the P.P.U.C. has approved the interconnection of the customer=s renewable energy system and the P.P.U.C. grid and shall not be interpreted to impose liability or approval by the P.P.U.C. for any part of the renewable energy system, its design, or its method of implementation. The technical standards imposed will be based solely on those necessary to ensure the safety of P.P.U.C. personnel and for the maintenance of P.P.U.C, power quality. Standards and technical requirements shall be consistent with existing technical practices for similar types of installations in the United States, Australia, or the European Union.
+	* A licensee shall inform the P.P.U.C. of any proposed technical changes to the renewable energy system that affects either the maximum power output or the components that provide the interconnection between the renewable energy system and the P.P.U.C. grid and will, under the licensing agreement, not make those changes without P.P.U.C. approval.
+	* The failure of a licensee to promptly inform the P.P.U.C. in writing of any technical changes to the renewable energy system that affects any of the above may, at the P.P.U.C. discretion, result in a fine of not more than two hundred dollars ($200).
+	* shall, at its own-expense, make available to each of its eligible customer generators who have installed a net metering system the meter (or set of meters) that is needed to determine the net flow of electricity both into and out of the electricity grid;"""
+    input_text_block = {
+        "text": [
+            "\n<li1>\n<Lbl>(d)<\\Lbl><LBody>shall, at its own expense, annually inspect grid-connected renewable energy installations to ensure that unauthorized changes have not been made and to ensure that the grid interconnection arrangements remain adequate for maintaining safety and power quality.<\\LBody>\\n",
+            "<Lbl>(e)<\\Lbl><LBody>shall not charge the customer any additional standby, capacity, interconnection, or other fee or charge that is greater than such fees charged to all members of that customer class; and<\\LBody>\\n",
+            "<Lbl>(f)<\\Lbl><LBody>may, at its own expense, and with the written consent of the customer, install one or more additional meters to monitor the flow of electricity in each direction. The additional metering shall be used only to provide the information necessary to accurately bill or credit the customer-generator or to collect renewable energy generating system performance information for research purposes.<\\LBody>\\n\\n<\\li1>\\n",
+        ],
+        "text_block_id": "p4_b96",
+        "coords": [
+            [44.00691223144531, 27.4151611328125],
+            [549.7152709960938, 27.4151611328125],
+            [44.00691223144531, 226.27561950683594],
+            [549.7152709960938, 226.27561950683594],
+        ],
+        "type": "list",
+        "path": ["Document", "L[4]", "LI[4]", "Lbl"],
+        "custom_attributes": {
+            "possibly_contiguous_with_same_page_context": True,
+            "paths": [
+                ["Document", "L[4]", "LI[4]", "Lbl"],
+                ["Document", "L[4]", "LI[4]", "LBody"],
+                ["Document", "L[4]", "LI[5]", "Lbl"],
+                ["Document", "L[4]", "LI[5]", "LBody"],
+                ["Document", "L[4]", "LI[6]", "Lbl"],
+                ["Document", "L[4]", "LI[6]", "LBody"],
+            ],
+            "text_block_ids": [
+                "p4_b96",
+                "p4_b97",
+                "p4_b98",
+                "p4_b99",
+                "p4_b100",
+                "p4_b101",
+            ],
+            "pretty_list_string": "* shall, at its own expense, annually inspect grid-connected renewable energy installations to ensure that unauthorized changes have not been made and to ensure that the grid interconnection arrangements remain adequate for maintaining safety and power quality.\\n\n* shall not charge the customer any additional standby, capacity, interconnection, or other fee or charge that is greater than such fees charged to all members of that customer class; and\\n\n* may, at its own expense, and with the written consent of the customer, install one or more additional meters to monitor the flow of electricity in each direction. The additional metering shall be used only to provide the information necessary to accurately bill or credit the customer-generator or to collect renewable energy generating system performance information for research purposes.\\n\\n\\n\n",
+            "original_list_text": [
+                ["(d)"],
+                [
+                    " shall, at its own expense, annually inspect grid-connected renewable energy installations to ensure that unauthorized changes have not been made and to ensure that the grid interconnection arrangements remain adequate for maintaining safety and power quality. "
+                ],
+                ["(e)"],
+                [
+                    " shall not charge the customer any additional standby, capacity, interconnection, or other fee or charge that is greater than such fees charged to all members of that customer class; and "
+                ],
+                ["(f)"],
+                [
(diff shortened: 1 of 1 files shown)


pipeline/processor/postprocessor.py
```diff
@@ -169,13 +169,15 @@ def merge_text_blocks(self, text_blocks: List[dict]) -> dict:
 
             # Append style metadata that will later be added to custom attributes for text block.
             if block_styling:
-                last_style_ix = len(merged_block_text[0].strip()) - 1
-                style_spans.append(
-                    {
+                # The last style index here is the index of the block with sentences joined.
+                last_style_ix = len(''.join((text for text in merged_block_text)).strip()) - 1
+                style_span = {
```
stefl14, line 174: This is a simple bug fix I noticed where I failed to account for the the length of the whole joined string in a list of strings, which was messing up some of the Spacy style indexing for style blocks, so needed an upstream fix.

pipeline/processor/postprocessor.py
```diff
@@ -308,7 +310,7 @@ def _format_semantic_lists(self, df: pd.DataFrame) -> List[List[str]]:
         df["type"] = df["path"].apply(lambda x: x[-1])
         # Sometimes label is called ExtraCharSpan, replace it with label
         df["type"] = df["type"].replace(
-            {"ExtraCharSpan": "Lbl", "ParagraphSpan": "LBody"}
+            {"ExtraCharSpan": "Lbl", "ParagraphSpan": "LBody", "Span": "Lbl"}
```
stefl14, line 313: Span elements complicate things a lot with Adobe and there's nothing general. But I noticed that assuming span elements are labels if they belong to list elements works in most cases. Another upstream fix.

pipeline/processor/postprocessor.py
```diff
@@ -404,10 +406,9 @@ def _update_contiguity_attributes(text_block: dict, new_attribute: str) -> dict:
 
         """
         new_custom_attributes = {new_attribute: True}
-        try:
-            if text_block["custom_attributes"] is not None:
-                text_block["custom_attributes"].update(new_custom_attributes)
-        except KeyError:
+        if text_block["custom_attributes"] is not None:
+            text_block["custom_attributes"].update(new_custom_attributes)
+        else:
```
stefl14, line 411: bug fix. Nothing to see here.

pipeline/reformat_adobe_output.py
```diff
@@ -33,7 +31,7 @@ def process(in_path, out_path):
             doc = text_styling_processor.process(doc, )
             doc = postprocessor.process(doc, f"{file.stem}.pdf")
             # Write to json file.
-            doc.save_json(out_path / file.stem)
+            doc.save_json(out_path / f"{file.stem}.json")
```
stefl14, line 34: was writing to text files, not json. Fixed.

pipeline/processor/postprocessor.py
```diff
@@ -576,11 +578,11 @@ def _group_list_elements(self, contents: dict) -> dict:
...
-                            else:
-                                text_block = self._update_contiguity_attributes(
-                                    text_block,
-                                    "possibly_contiguous_with_same_page_context",
-                                )
+                        else:
+                            text_block = self._update_contiguity_attributes(
+                                text_block,
+                                "possibly_contiguous_with_same_page_context",
+                            )
```
stefl14, line 585: Noticed a mistake where I wasn't appending this metadata at the right tabulation level, so it sometimes wasn't being included. It's not really being used in the code as is though, so not really anything to worry about.

search-index/text2embeddings.py
```diff
@@ -18,10 +19,108 @@
 logger = get_logger(__name__)
 
 
+def get_text_from_list(text_block: dict, prev_processed_text_block: dict) -> str:
```
stefl14, line 22: Tried a bunch of stuff here, but eventually settled on this ridiculously simple function as more complex implementations just created more headaches than they solved. See PR comments for more context.

search-index/text2embeddings.py
```diff
@@ -18,10 +19,108 @@
...
+    Args:
+        data (str): string to delete indices from.
+        indices (List[int]): list of indices to delete.
+    Returns:
+        str: string with indices deleted.
+    """
+    return "".join([char for idx, char in enumerate(data) if idx not in indices])
+
+
+def get_text_from_merged_block(text_block: dict) -> str:
```
stefl14, line 60: This is where the bulk of the interesting stuff is, though still super simple.

kdutia:
@stefl14 I've noticed the following warnings in `test_pipeline`. Usually these are because you need to set the string to be a raw string (`r"the string"` or `fr"the {value}"`).

```
=============================== warnings summary ===============================
processor/postprocessor.py:328
  /app/processor/postprocessor.py:328: DeprecationWarning: invalid escape sequence \L
    label_string = f"<Lbl>{text}<\Lbl>"

processor/postprocessor.py:340
  /app/processor/postprocessor.py:340: DeprecationWarning: invalid escape sequence \L
    new_string += f"<LBody>{text}<\LBody>\n" + f"\n<\li{list_number}>\n"

processor/postprocessor.py:340
  /app/processor/postprocessor.py:340: DeprecationWarning: invalid escape sequence \l
    new_string += f"<LBody>{text}<\LBody>\n" + f"\n<\li{list_number}>\n"

processor/postprocessor.py:342
  /app/processor/postprocessor.py:342: DeprecationWarning: invalid escape sequence \L
    new_string += f"<LBody>{text}<\LBody>\n"

processor/postprocessor.py:429
  /app/processor/postprocessor.py:429: DeprecationWarning: invalid escape sequence \d
    df["block_num"] = df["text_block_id"].str.extract("b(\d+)").astype(int)

processor/postprocessor.py:457
  /app/processor/postprocessor.py:457: DeprecationWarning: invalid escape sequence \d
    df["text_block_id"].str.split("_").str[0].str.extract("(\d+)").astype(int)
```

pipeline/processor/postprocessor.py
kdutia: 🙌🏼

search-index/text2embeddings.py
```diff
@@ -18,10 +19,108 @@
...
+    Args:
+        text_block (dict): text block dict.
+        prev_processed_text_block (dict): previous text block dict.
+
+    Returns:
+        str: Formatted list text block that is useful for indexing semantics.
+    """
+    text = text_block["custom_attributes"]["pretty_list_string"]
+    text = text.strip()
+    text = text.replace("*", "")  # Remove bullet points.
```
kdutia, line 36: I suggest we leave bullet points in as we'll display this processed text to the user on the search results page. If we decide it's better for NLP purposes to remove them, we can remove them for search purposes only in Elasticsearch using a [character filter](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pattern-replace-charfilter.html) and a [field](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html).

search-index/text2embeddings.py
```diff
@@ -18,10 +19,108 @@
...
+    Args:
+        text_block: A block of text that has been merged.
+
+    Returns:
+        str: The text block with styling that is not related to semantics removed.
+
+    """
+    # Text output with no processing.
+    text_output = (
+        "\n".join(text_block["text"]).strip(),
```
kdutia, line 73: I think we want to actually `" ".join()` strings here so that displaying them in the UI doesn't enforce arbitrary line breaks. We should still newline join where newlines are part of the semantics, e.g. for lists or paragraph breaks. This is my mistake from the original method that I've spotted since. What do you think?

search-index/text2embeddings.py
```diff
@@ -31,12 +130,18 @@ def get_text_from_document_dict(document: dict) -> List[Tuple[str, str]]:
...
         for text_block in page["text_blocks"]:
-            text_output.append(
-                ("\n".join(text_block["text"]).strip(), text_block["text_block_id"])
-            )
+            if text_block["type"] == "merged_text_block":
+                text = get_text_from_merged_block(text_block)
+            elif text_block["type"] == "list":
+                text = get_text_from_list(text_block, prev_block_processed_text)
+            else:
+                text = "\n".join(text_block["text"]).strip()
```
kdutia, line 141: Same here with `" ".join` as in above comment.

search-index/text2embeddings.py
```diff
@@ -18,10 +19,108 @@
...
+                # be used by language models.
+                delete_substr = text_output[0][remove_start : remove_end + 1]
+                if re.match(r"(st|nd|rd|th)$", delete_substr):
+                    continue  # don't remove
+                delete_indices.extend([remove_start, remove_end])
+
+        # The indices are relative to the whole block, not just elements in the block.
+        # Remove the indices from deleted_indices from text_output.
+        if delete_indices:  # There will only be indices to delete in superscript case.
+            text_output_amended = delete_string_indices(text_output[0], delete_indices)
```
kdutia, line 96: It seems like you're passing start, end pairs into this function but the function actually deletes characters by index? Should the `delete_indices.extend` line add a range of indices from start to end instead?

search-index/test/test_text2embeddings.py
```diff
@@ -0,0 +1,83 @@
...
+        "coords": [
+            [85.08000183105469, 207.0596923828125],
+            [555.9367065429688, 207.0596923828125],
+            [85.08000183105469, 278.4237060546875],
+            [555.9367065429688, 278.4237060546875],
+        ],
+        "type": "merged_text_block",
+        "path": ["Document", "P[40]"],
+        "custom_attributes": {
+            "styleSpans": [{"style": "superscript", "start_idx": 144, "end_idx": 144}]
```
kdutia, line 21: Only just noticed this but the start_idx and end_idx behaviour is a bit un-pythonic here - you'd expect the representation of only index 144 to be (144,145). Happy to leave it like this as it looks like other code might depend on it, as long as a note about this behaviour is in the relevant docstring.

kdutia:
Looks good, i've left some pretty minor comments. Are you planning on adding a test for the `get_text_from_list` function?

search-index/text2embeddings.py
stefl14: Good spot. Since I was only really dealing with superscripts and mostly cases of a single superscript, hadn't spotted it!

search-index/text2embeddings.py
```diff
@@ -10,6 +10,7 @@
 import pandas as pd
 import click
 import re
+import codecs
```
stefl14, line 13: hack

stefl14:
Merged too soon.

PR #444, opened 2022-04-07
Add data and make command

kdutia:
Example files created using the text2embedding CLI that can be used to populate Opensearch via the command `make populate_opensearch_test`. There are 5,000 passages across 6 documents in this data sample.

I've made this new `make` command dependent on `make run_loader` to prevent the most obvious mistake I can see with running this command, but this means it takes an extra minute even if you've already got the documents loaded into postgres as it checks that this is the case.

@joel-wright happy for you to merge if you're happy with it.


joel-wright (approved):
This is really useful for setting up a working system! We might want to think carefully about the data we might want for integration testing in the future, but this is a great start.

PR #445, opened 2022-04-07
Feat 391 text transformation function doover

stefl14:


kdutia (changes requested):
@stefl14 judging by the original PR #435, there are still quite a few changes missing from here that were in the original PR.

PR #449, opened 2022-04-08
Add Search Endpoint

joel-wright:
This is still WIP, just opening it up for comments.


backend/app/api/api_v1/routers/tests/test_search.py
```diff
@@ -0,0 +1,171 @@
...
+    response = client.post(
+        "/api/v1/searches",
+        json={"query_string": "disaster", "exact_match": False},
+        headers=user_token_headers,
+    )
+    end = time.time()
+
+    processed_response_time_ms = 1000 * (end - start)
+    assert response.status_code == 200
+    assert 0 < response.json()["query_time_ms"] < processed_response_time_ms
```
opyate, line 171: Is it more a case of checking this attribute is in the response, rather than checking rough accuracy?
Or perhaps, if you have a "really fast computer" :-)

```
assert 0 < response.json()["query_time_ms"] <= processed_response_time_ms
```

?

joel-wright: Some of the responses I've seen have been in the low single digit ms, so while this isn't a great test,,, it did at least check that some time had passed and it wasn't obviously wrong... I'll think about how to make it better!

backend/app/db/schemas/search.py
```diff
@@ -0,0 +1,115 @@
...
+    offset: int = 0
+
+
+class SearchResponseDocumentPassage(BaseModel):
+    """A Document passage match returned by the search API endpoint."""
+
+    text: str
+    text_block_id: str
+    text_block_page: int
+    text_block_coords: List[List[int]]
```
kdutia, line 47: I think the type of coords should be `List[List[float]]`

backend/app/db/schemas/search.py
```diff
@@ -0,0 +1,115 @@
...
+    for_search_action_description: str
+
+
+class OpenSearchResponsePassageMatch(OpenSearchResponseMatchBase):
+    """Describes matches returned by OpenSearch on Document passage."""
+
+    text: str
+    text_block_id: str
+    text_block_page: int
+    text_block_coords: List[List[int]]
```
kdutia, line 115: The type here should also be `List[List[float]]` - see my last comment

backend/app/db/schemas/search.py
```diff
@@ -0,0 +1,115 @@
...
+    limit: int = 10  # TODO: decide on default
+    offset: int = 0
+
+
+class SearchResponseDocumentPassage(BaseModel):
+    """A Document passage match returned by the search API endpoint."""
+
+    text: str
+    text_block_id: str
+    text_block_page: int
```
kdutia, line 46: Note that the page number returned here is 0-indexed. This is maybe counter-intuitive: the front end would likely want to add 1 to this value for it to make sense to users. We should either add 1 in the API, or write in the API docs that it's 0-indexed?

backend/app/core/search.py
```diff
@@ -0,0 +1,466 @@
...
+_DEFAULT_INNER_PRODUCT_THRESHOLD: float = 70.0
+_DEFAULT_MAX_DOC_COUNT: int = 100
+_DEFAULT_MAX_PASSAGES_PER_DOC: int = 10
+_DEFAULT_N_PASSAGES_TO_SAMPLE_PER_SHARD: int = 5000
+_DEFAULT_NAME_BOOST: int = 100
+_DEFAULT_DESCRIPTION_BOOST: int = 100
+_DEFAULT_OPENSEARCH_INDEX = "navigator"
+_DEFAULT_OPENSEARCH_REQUEST_TIMEOUT = 30
+_DEFAULT_OPENSEARCH_PREFERENCE = "prototype_user"  # TODO: document what this means
+_DEFAULT_OPENSEARCH_EMBEDDING_DIM = 768
```
kdutia, line 28: This parameter isn't needed at search time as the index mapping is designed to only accept embeddings with this length. If we wanted to do a check that the query embedding length matches the length Opensearch expects in future, we'd probably query the Opensearch index to get this value and then compare it to `emb.shape[0]` where `emb = _ENCODER.encode("query")`.

backend/app/core/search.py
```diff
@@ -0,0 +1,466 @@
...
+    url: str
+    username: str
+    password: str
+    index_name: str = _DEFAULT_OPENSEARCH_INDEX
+    embedding_dim: int = _DEFAULT_OPENSEARCH_EMBEDDING_DIM
+    request_timeout: int = _DEFAULT_OPENSEARCH_REQUEST_TIMEOUT
+    preference: str = _DEFAULT_OPENSEARCH_PREFERENCE
+    use_ssl: bool = False
+    verify_certs: bool = False
+    ssl_show_warnings: bool = False
```
kdutia, line 53: The values of these parameters will probably change when we connect to the managed service as it uses SSL. I guess we want them as environment variables so we can still run tests against unsecured Opensearch instances in dev/test?

backend/app/core/search.py
```diff
@@ -0,0 +1,466 @@
...
+
+    name_boost: int = _DEFAULT_NAME_BOOST
+    description_boost: int = _DEFAULT_DESCRIPTION_BOOST
+    lucene_threshold: float = _innerproduct_threshold_to_lucene_threshold(
+        _DEFAULT_INNER_PRODUCT_THRESHOLD
+    )  # TODO: tune me separately for descriptions?
+    max_doc_count: int = _DEFAULT_MAX_DOC_COUNT
+    max_passages_per_doc: int = _DEFAULT_MAX_PASSAGES_PER_DOC
+    n_passages_to_sample_per_shard = _DEFAULT_N_PASSAGES_TO_SAMPLE_PER_SHARD
+    k = 10000  # TODO: tune me
```
kdutia, line 169: Could we assign this at the top as with `max_passages_per_doc` etc, as its value will probably need to be tweaked in the same thought process as some of the others? Maybe `_DEFAULT_KNN_K_VALUE`?

backend/app/api/api_v1/routers/tests/test_search.py
```diff
@@ -0,0 +1,289 @@
...
+        "/api/v1/searches",
+        json={"query_string": "", "exact_match": False},
+        headers=user_token_headers,
+    )
+    assert response.status_code == 200
+    assert response.json()["hits"] == 0
+
+    response = client.post(
+        "/api/v1/searches",
+        json={"query_string": "", "exact_match": False},
```
kdutia, line 285: Should this be `"exact_match": True`?

kdutia:
@joel-wright overall looks good 👍🏼. I've added some minor comments above but apart from those happy for you to merge.

backend/app/db/schemas/search.py
joel-wright: I think I might go a bit further and actually define some types, but I'll make sure to define `Coord = Tuple[float, float`]

backend/app/db/schemas/search.py
joel-wright: let's just add one in the processing - the backend code is already processing everything so it shouldn't make a massive difference

backend/app/api/api_v1/routers/tests/test_search.py
joel-wright: it really should :grin:

backend/app/db/schemas/search.py
kdutia: by processing do you mean during indexing or during searching?

joel-wright: SOrry, yes, I meant search time. When we process the result of the search from OpenSearch into the API response.

.env.example
```diff
@@ -8,6 +8,27 @@ OPENSEARCH_USER=admin
...
+OPENSEARCH_INDEX_N_PASSAGES_TO_SAMPLE_PER_SHARD=5000
+OPENSEARCH_INDEX_NAME_BOOST=100
+OPENSEARCH_INDEX_DESCRIPTION_BOOST=10
+OPENSEARCH_INDEX_NAME_KEY=for_search_action_name
+OPENSEARCH_INDEX_DESCRIPTION_KEY=for_search_action_description
+OPENSEARCH_INDEX_DESCRIPTION_EMBEDDING_KEY=action_description_embedding
+OPENSEARCH_INDEX_INDEX_KEY=action_name_and_id
+OPENSEARCH_INDEX_TEXT_BLOCK_KEY=text_block_id
+OPENSEARCH_INDEX_ENCODER=msmarco-distilbert-dot-v5
+OPENSEARCH_INDEX_EMBEDDING_DIM=768
```
kdutia, line 31: this can be deleted ☺️

joel-wright: I added that to the env file because it was really the only part of the index config I was aware of that wasn't configurable. it's not used by any of the backend code, but I was hoping this might be the start of shared config between the `search-index` code and the `backend` ?

PR #459, opened 2022-04-12
Add Search Sort Ordering

joel-wright:
Sorting now supported ascending/descending according to:
- score
- date
- title


backend/app/api/api_v1/routers/tests/test_search.py
```diff
@@ -140,11 +137,126 @@ def test_multiple_filters(
     } in query_body["query"]["bool"]["filter"]
 
 
-def test_result_order(test_opensearch, monkeypatch, client, user_token_headers):
+@pytest.mark.parametrize("order", [SortOrder.DESCENDING, SortOrder.DESCENDING])
```
kdutia, line 140: I think this should be `[SortOrder.DESCENDING, SortOrder.ASCENDING]`

kdutia, line 140: Do we need `DESCENDING` twice here?

backend/app/api/api_v1/routers/tests/test_search.py
```diff
@@ -140,11 +137,126 @@ def test_multiple_filters(
...
+        new_s = d["top_hit"]["value"]
+        if s is not None:
+            if order == SortOrder.DESCENDING:
+                assert new_s <= s
+            if order == SortOrder.ASCENDING:
+                assert new_s >= s
+        s = new_s
+
+
+@pytest.mark.parametrize("order", [SortOrder.DESCENDING, SortOrder.DESCENDING])
```
kdutia, line 171: Think this should be ASCENDING, DESCENDING. Same with `test_result_order_title` below (but not the score one).

backend/app/db/schemas/search.py
```diff
@@ -35,7 +34,7 @@ class SearchRequestBody(BaseModel):
     year_range: Optional[Tuple[Optional[int], Optional[int]]] = None
 
     sort_field: SortField = SortField.DATE
```
kdutia, line 36: Does this mean that the default sort order of search results is date descending? It should be score descending, right?

kdutia:
Looks good! I couldn't leave a comment inline as the line wasn't changed in this PR, but I think the `stats` clause in the Opensearch query could be amended as follows, as that clause's only purpose is to enable sorting by date.

``` diff
"action_date": {
    "stats": {
-        "field": "action_date",
+        "field": _SORT_FIELD_MAP[SortField.DATE],
    },
},
```

backend/app/api/api_v1/routers/tests/test_search.py
joel-wright: :man_facepalming:

backend/app/db/schemas/search.py
joel-wright: I really should keep a more careful eye on autocomplete

PR #465, opened 2022-04-13
Add service healthchecks

joel-wright:
Remove sleep hack, add service health checks for slow starting services & add dependencies on service health for backend/index tasks


opyate (approved):
ship it

PR #473, opened 2022-04-14
Feat 109 run indexing s3

stefl14:
Relatively small PR in the end as most of the work turned out to be getting it to run on a GPU VM. See commits for details.

outputs and logs stored at:

 https://s3.console.aws.amazon.com/s3/buckets/cpr-cclw-cpd-docs-temp?region=eu-west-2&prefix=embeddings_update_14_04_2022/&showversions=false
 
Quite a significant issue flagged from the log files can be found at:

https://app.zenhub.com/workspaces/climate-policy-radar-613867967a1a9200123c9aad/issues/climatepolicyradar/website/110

### DIFF
--- a/search-index/text2embeddings.py
+++ b/search-index/text2embeddings.py
@@ -59,8 +59,11 @@ def get_text_from_list(text_block: dict, prev_processed_text_block: dict) -> str
     #  when, for example, there are random elements between pages that the postprocessors haven't been able to catch.
     #  But I've ignored this for now because it seems that the postprocessing in postprocessor.py has dealt with the
     #  vast majority of edge cases (as indicated by the rarity of certain debugging metadata).
-    text = prev_processed_text_block + "\n" + text
-    return text
+    if prev_processed_text_block:
+        text = prev_processed_text_block + "\n" + text
+        return text
+    else:
+        return text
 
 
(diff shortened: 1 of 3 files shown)


search-index/text2embeddings.py
```diff
@@ -323,6 +367,22 @@ def run_cli(
 
     logger.info(f"Saved embeddings and IDs for text and descriptions to {output_dir}")
 
+    if s3:
+        logger.info(f"Writing embeddings to S3 bucket {output_dir}")
+        output_dir.upload_from(embs_output_path, force_overwrite_to_cloud=True)
```
kdutia, line 372: What's the reason for setting `force_overwrite_to_cloud=True`? Timestamping files should ensure their uniqueness so it could be quite useful to find out if a file is going to be overwritten, as this would mean timestamping is probably broken.

stefl14: Nice spot. I think this is because I I was debugging with a subset of 1 file for speed but the time string is only 1m accuracy, so I ended up running the script twice within a minute and wrote this in to avoid the bugs, copy and pasted it and forgot to remove! Fixed.

kdutia (approved):
lgtm 🎉

PR #475, opened 2022-04-19
Feat 400 - move origin to bottom left for PDF parsers

kdutia:
Replaces #468 as it was easier to start the work again on a new branch. I've reverted the Adobe parser to use the coordinates Adobe returns (bottom left origin) and flipped the ones that pdfalto returns (top left origin). There were also some mypy exceptions to clean up.

Comments about Adobe Embed from #468:

---
**Update: I've added these details [to Notion](https://www.notion.so/climatepolicyradar/Adobe-PDF-Embed-API-bounding-box-highlighting-4d8a60d2f00b44558cb25358ce0d333d) too**

This PR adds a new postprocessor which flips the coordinate system used for text block coordinates so that the origin is on the bottom left of each page, as this is what's required by the Adobe Embed API. Note there are some other small code changes but these were just to get the pre-commit checks to pass.

I've also forked a codepen showing how we can add multiple bounding box annotations to one PDF based on bounding boxes.

* [codepen with working example](https://codepen.io/kdutia/pen/QWaVVdO)
* [Adobe Embed API annotation docs](https://developer.adobe.com/document-services/docs/overview/pdf-embed-api/howtos_comments/#annotation-schema)
* [helpful article](https://medium.com/adobetech/adding-annotations-to-a-pdf-using-adobe-pdf-embed-api-fb6f85da4c02) (view in private browsing to bypass medium's paywall)

**Screenshot of codepen output.** Note the sidebar is optional and can be disabled by setting `showCommentsPanel: false`

image

I have the following **outstanding question**:
* the Embed API requires coordinates in two different formats: a bounding box (`[Xmin, Ymin, Xmax, Ymax]`) and a set of rectangles to draw within that bounding box, which for our purpose is the same rectangle but needs to be defined using the format `[Xmin, Ymin, Xmax, Ymin, Xmax, Ymax, Xmin, Ymax]`. At the moment through the API we provide the latter but not the former. Should we also provide the simplified coordinates through Opensearch, or will this be easy enough to do on the front end?

### DIFF
--- a/pipeline/extract/extract.py
+++ b/pipeline/extract/extract.py
@@ -168,32 +168,41 @@ class DocumentEmbeddedTextExtractor(DocumentTextExtractor):
 
         return xml_output_path
 
-    def _get_text_block_coords(self, text_block: Element) -> List[Tuple[float, float]]:
+    def _get_text_block_coords(
+        self, text_block: Element, page_height: float
+    ) -> List[Tuple[float, float]]:
         """Get the coordinates of a text block element.
 
         Fetches the x, y, width and height coordinates from an alto xml text block element
-        and returns as a list of x, y pairs.
+        and returns as a list of x, y pairs, with the origin at the bottom left of the page.
+
+        pdfalto provides [x0, y0, x1, y1] with origin at top left. We flip y coordinates so that the
+        origin is at the bottom left.
 
         Args:
-            text_block: Element representing a text block in the XML tree
+            text_block (Element): Element representing a text block in the XML tree
+            page_height (float): height of the page
 
         Returns:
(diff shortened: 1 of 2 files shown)


joel-wright (approved):
LGTM. I did notice a pattern in the post-processing code that we should look at as a future tech debt ticket; converting a copy of a dataclass/pydantic model into a dict which we then modify in-place means we don't get any benefit from the type checker... not something for pre-alpha though, probably a task to take on while we're getting the pipeline code running as a repeatable workflow :)

kdutia:
@joel-wright I've raised #477 with a link to this PR so we can come back to that code issue later

PR #483, opened 2022-04-21
Deployment

opyate:
Stuff to get Pulumi-based deployment working.

### DIFF
--- a/backend/app/core/config.py
+++ b/backend/app/core/config.py
@@ -59,5 +59,5 @@ OPENSEARCH_INDEX_TEXT_BLOCK_KEY: str = os.getenv(
     "OPENSEARCH_INDEX_TEXT_BLOCK_KEY", "text_block_id"
 )
 OPENSEARCH_INDEX_ENCODER: str = os.getenv(
-    "OPENSEARCH_INDEX_ENCODER", "msmarco-distilbert-dot-v5"
+    "OPENSEARCH_INDEX_ENCODER", "sentence-transformers/msmarco-distilbert-dot-v5"
 )
--- a/infra/cpr/plumbing/main.py
+++ b/infra/cpr/plumbing/main.py
@@ -1,6 +1,6 @@
 """Infra-as-code for CPR stack."""
 import json
-import os
+import pathlib
 import pulumi
 import pulumi_aws as aws
 
@@ -67,7 +67,7 @@ class Plumbing:
             most_recent=True,
         )
 
-        stack_name = os.environ["PULUMI_STACK"]
+        stack_name = pulumi.get_stack()
 
         keypair = aws.ec2.KeyPair(
             "bastion-ssh-key",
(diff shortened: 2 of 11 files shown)


backend/app/core/config.py
```diff
@@ -59,5 +59,5 @@
     "OPENSEARCH_INDEX_TEXT_BLOCK_KEY", "text_block_id"
 )
 OPENSEARCH_INDEX_ENCODER: str = os.getenv(
-    "OPENSEARCH_INDEX_ENCODER", "msmarco-distilbert-dot-v5"
+    "OPENSEARCH_INDEX_ENCODER", "sentence-transformers/msmarco-distilbert-dot-v5"
```
opyate, line 62: Wondering if we need to bring `OPENSEARCH_INDEX_ENCODER` in here somehow and in the Dockerfile, but then should it really be configurable in the first place? E.g. Could it ever be set to a value that's incompatible with the code?

infra/cpr/plumbing/main.py
```diff
@@ -25,10 +25,62 @@ class Plumbing:
...
+            owners=["137112412989"],  # This owner ID is Amazon
+            most_recent=True,
+        )
+
+        stack_name = pulumi.get_stack()
+
+        keypair = aws.ec2.KeyPair(
+            "bastion-ssh-key",
+            key_name=f"bastion-{stack_name}-ssh-key",
+            public_key="ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA14swrmHESs/dE8+nhj8twLPpbq//FHAVw7fXEQIvflJ+03yb6ostNLcXNroenoKzfRAwqVx0BRgYB0RKoBP5yIL7K5UkzxnWHEpIj3oC+kIrR1kn8LlsGa1aP05BYmZvsBXFzBjiBie3vhOTGOiEVEzqxdJLZExa6HIeoarKukAi94uQNDFZbnBNqdjRdgIPUdDHYlL5abSKC+yTS2+L96y6Dzt47i+rkn2rZzld1nYgRNLfAtvooIPZ60IfCnhRGPGuPA+nyYLtgRYVKCfC5yZ6QFUp0WH9NjE7PJMY2UX6w465bgbWlVtu8Ue687oWNyJf3r39S73gmpHOUHMkuw== opyate@gmail.com",
```
opyate, line 75: Oh, just realised this might be a problem for bastion usage. I couldn't find a way to incorporate more public keys in here. Maybe one for a near-future task?
You can always just temporarily plonk your own public key in here if you need to use the bastion.

infra/cpr/plumbing/main.py
```diff
@@ -180,10 +226,10 @@ def __init__(self):
 
         # TODO A gateway and routing table are needed to allow the VPC to communicate with the Internet.
         # Once created, we associate the routing table with our VPC.
-        """
+        _comment = """
```
opyate, line 229: I think this was to shut up a linter. Please disregard.

infra/cpr/plumbing/main.py
opyate: Or perhaps we set up a shared public/private key combo, and put it in our respective .ssh folders.

joel-wright: would be good to work out how we use AWS creds to access this - sharing keys is a pain in the long run & is a pain if someone needs to have their access revoked.

opyate: https://github.com/climatepolicyradar/navigator/issues/484

joel-wright (approved):
This gets the deployment working, and I'll raise follow-up tickets to address the ssh key and machine instance questions.

PR #489, opened 2022-04-22
Feat 432 opensearch service

kdutia:
In this PR:
* indexing changes to speed up aggregation performance (eagerly load global ordinals; disable refresh interval during indexing to ensure 1 segment per shard)
* set 'preference' parameter for search based on user ID. This ensures that a user's prior searches are cached.
* notebook to run query tests on index.
* push notebook with tests for three-node, 2 replica setup
* add index warmup request once indexing is finished
* script to create cloudwatch alarms

See notes on the final production config settled on [in Notion](https://www.notion.so/climatepolicyradar/Opensearch-production-config-ce6f5c5d9c6f4c218a04f86256a3461e#01f833fa0bad42eb9ce80bba488a97f8).


kdutia:
Ready for review. The CloudWatch alarms script has run successfully, and I'll check that the index warmup runs successfully as part of the indexing process this afternoon.

search-index/index_data.py
```diff
@@ -225,16 +225,30 @@ def run_cli(
...
     )
 
     opensearch.delete_and_create_index()
+    # We disable index refreshes during indexing to speed up the indexing process,
+    # and to ensure only 1 segment is created per shard. This also speeds up KNN
+    # queries and aggregations according to the Opensearch and Elasticsearch docs.
+    opensearch.set_index_refresh_interval(-1, timeout=60)
     opensearch.bulk_index(actions=doc_generator)
 
+    # TODO: we wrap this in a try/except block because for now because sometimes it times out
```
stefl14, line 246: Is the timeout here just a flaky error that occurs occasionally? I'm ignorant of what the opensearch py method is doing under the hood here, but if it's not pinging the server intelligently according to the timeout interval it might be better to wrap this in a retry loop if it's likely to guarantee success after a few tries, in case we forget to reset the interval (how important is this?). This depends on how the method is working under the hood ofc, so may be redundant comment.

stefl14 (commented):
LGTM,  few comments, mostly for my own clarity really!

.env.example
```diff
@@ -9,7 +9,6 @@ OPENSEARCH_PASSWORD=admin
 OPENSEARCH_URL=http://opensearch-node1:9200
 OPENSEARCH_INDEX=navigator
 OPENSEARCH_REQUEST_TIMEOUT=30
-OPENSEARCH_PREFERENCE=prototype_user  # TODO: document what this means
```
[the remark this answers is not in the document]
kdutia: [Here's the relevant Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-search.html#search-preference) - it basically describes that queries with a fixed preference parameter will always follow the same node and shard routing logic, which means you benefit from query caching. 

By making this change to replace a fixed value with the user ID ([this commit](https://github.com/climatepolicyradar/navigator/pull/489/commits/b407b251f25dfefde33c5353751f4fc9df8732b1)) we gain a bit of efficiency as Opensearch has a bit more freedom to route queries according to load on each of the nodes. I got this idea from [optimising search speed](https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-search-speed.html#preference-cache-optimization) page of the Elasticsearch docs which is worth a quick read.

search-index/index_data.py
kdutia: Good point. I chose to wrap it in a try/except for now as we don't care what the refresh interval is, as we're not incrementally adding documents. 

If(/when) we move to a DAG-managed indexer I imagine the steps of disabling the index refresh, indexing, and re-enabling the index refresh would be separate processes that we could track and re-run if needed.

Relevant part of the Elasticsearch docs: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-refresh.html

.env.example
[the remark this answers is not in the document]
stefl14: That's clarifying. Cheers.

backend/app/api/api_v1/routers/search.py
```diff
@@ -31,4 +31,5 @@ def search_documents(
     return _OPENSEARCH_CONNECTION.query(
         search_request_body=search_body,
         opensearch_internal_config=_OPENSEARCH_INDEX_CONFIG,
-    )
+        preference=str(current_user.id)
+    )
```
joel-wright, line 35: this is weird, python files should end in a newline character... I would have thought flake8 would complain about that. no need to fix it, it'll be fixed automatically later.

joel-wright (approved):
Looks good, just noted one oddity that means I need to check the linting config :)

PR #491, opened 2022-04-25
Add facilities to merge in style blocks that Adobe cannot detect.

stefl14:
Since Adobe fails to detect some styling elements (italics, but also sometimes other elements), these elements have different paths and so are not included by simple "merge duplicated path" logic.

- This PR solves the issue by assuming that Intervening blocks between blocks with the same path are also part of those semantics.
- Unfortunately, this doesn't allow us to identify the style type (e.g. whether the previously missing semantics were bold, for example). Instead, I assign the last element of the path as the style type (usually Span). This is potentially helpful for debugging
purposes, allowing us to see the scope of the problem if we want a better fix later.
-  In general, this limitation won't affect effect semantic content.

### DIFF
--- a/pipeline/processor/postprocessor.py
+++ b/pipeline/processor/postprocessor.py
@@ -4,7 +4,7 @@ from abc import ABC, abstractmethod
 from collections import Counter
 from collections import defaultdict
 from copy import deepcopy
-from typing import List, Dict, Optional, Sequence
+from typing import List, Dict, Optional, Sequence, Tuple
 
 import pandas as pd
 from english_words import english_words_set
@@ -134,6 +134,8 @@ class AdobeTextStylingPostProcessor(PostProcessor):
             return "underline"
         elif text_block["custom_attributes"].get("TextPosition") == "Sup":
             return "superscript"
+        elif "UnknownStyle" in text_block["custom_attributes"]:
+            return text_block["custom_attributes"].get("UnknownStyle")
         else:
             return None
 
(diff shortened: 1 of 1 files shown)


pipeline/processor/postprocessor.py
```diff
@@ -196,6 +198,67 @@ def merge_text_blocks(self, text_blocks: List[dict]) -> dict:
...
+        for i, j in enumerate(all_paths):
+            if path == j:
+                duplicated_path_indices.append(i)
+        merge_path_indices = list(
+            range(duplicated_path_indices[0], duplicated_path_indices[-1] + 1)
+        )
+        return merge_path_indices, duplicated_path_indices
+
+    @staticmethod
+    def _update_style_metadata(
```
stefl14, line 237: This function is not strictly necessary. It's just a helper util for grabbing metadata on the path type of the newly included semantics. Unfortunately, this won't be "Bold", "Underline" etc, but is more likely to be "span", which may be somewhat helpful for visibility and debugging down the line.

pipeline/processor/postprocessor.py
```diff
@@ -227,7 +291,27 @@ def process(self, document: Document, filename=None) -> Document:
             duplicated_paths = [
                 path for path, count in path_counts.items() if count > 1
             ]
+
```
stefl14, line 297: One thing to not here is that the following path is duplicated, which isn't especially pythonic and could be refactored.

```for path in duplicated paths```

However, it was the quickest fix for the fact that the iterative merging process in merge_text_blocks deletes indices, so the indices calculated in get_paths_to_merge become stale throughout the loop. The solution was therefore to 
get the indices in the first loop, and perform merging in the second.

kdutia (approved):

PR #492, opened 2022-04-25
Feature-437 Extra filter lookups

joel-wright:
Adds:
- lookups for filter fields
- a mapping dictionary from lookup names to opensearch fields
- missing category table that came out of extra discussions
- an (experimental) transformation for lookups that point to self-referencing tables in the database to build a layered tree object as output, rather than just a list with foreign key references.


backend/app/api/api_v1/routers/lookups.py
```diff
@@ -16,15 +24,44 @@ def table_to_json(
...
+def tree_table_to_json(
+    table: Base,
+    db: SessionLocal,  # type: ignore
+) -> List[Dict]:
+    json_out = []
+    child_list_map: Dict[int, Any] = {}
+
+    for row in db.query(table).all():
+        row_object = {
+            col.name: getattr(row, col.name) for col in row.__table__.columns
```
opyate, line 44: If `getattr(row, col.name)` returns a date, it won't serialise. Maybe check for date/datetime instance, and use `.isoformat()`?
Alternatively, I'm doing something similar in a previous PR, and deferring the decision to use something like Marshmallow a bit later.

PR #507, opened 2022-04-29
Feat 470 post all metadata to api

opyate:


backend/app/api/api_v1/routers/tests/test_search.py
```diff
@@ -8,9 +8,8 @@
 from app.core.search import _FILTER_FIELD_MAP
 
 
-def test_simple_pagination(
-    test_opensearch, monkeypatch, client, user_token_headers
-):
+@pytest.mark.search
```
joel-wright, line 11: nice

backend/app/core/service/loader.py
```diff
@@ -0,0 +1,150 @@
...
+        db_event = Event(
+            document_id=db_document.id,
+            name=event.name,
+            description=event.description,
+        )
+        db.add(db_event)
+
+    # sectors
+    for meta in document_with_metadata.sectors:
+        insert_stmt = insert(Sector).values(
```
joel-wright, line 68: These inserts are a little surprising, I thought we were following a pattern of making sure all valid values were already present in the tables and we would use the uniqueness constraint to look them up...?

backend/app/core/service/loader.py
```diff
@@ -0,0 +1,150 @@
...
+            # TODO potentially use meta_id as parent_id
+            db_bridge = DocumentSector(
+                sector_id=meta_id,
+                document_id=db_document.id,
+            )
+            db.add(db_bridge)
+
+    # instruments
+    for meta in document_with_metadata.instruments:
+        insert_stmt = insert(Instrument).values(
```
joel-wright, line 87: similar to above re. insert

backend/app/core/service/loader.py
```diff
@@ -0,0 +1,150 @@
...
+            # TODO potentially use meta_id as parent_id
+            db_bridge = DocumentInstrument(
+                instrument_id=meta_id,
+                document_id=db_document.id,
+            )
+            db.add(db_bridge)
+
+    # hazards
+    for meta in document_with_metadata.hazards:
+        insert_stmt = insert(Hazard).values(
```
joel-wright, line 106: similar to above re. insert

backend/app/core/service/loader.py
```diff
@@ -0,0 +1,150 @@
...
+            meta_id = return_value.inserted_primary_key[0]
+            db_bridge = DocumentHazard(
+                hazard_id=meta_id,
+                document_id=db_document.id,
+            )
+            db.add(db_bridge)
+
+    # responses
+    for meta in document_with_metadata.responses:
+        insert_stmt = insert(Response).values(
```
joel-wright, line 122: similar to above re. insert

backend/app/core/service/loader.py
```diff
@@ -0,0 +1,150 @@
...
+            meta_id = return_value.inserted_primary_key[0]
+            db_bridge = DocumentResponse(
+                response_id=meta_id,
+                document_id=db_document.id,
+            )
+            db.add(db_bridge)
+
+    # frameworks
+    for meta in document_with_metadata.frameworks:
+        insert_stmt = insert(Framework).values(
```
joel-wright, line 138: similar to above re. insert

loader/app/service/tree_parser.py
```diff
@@ -0,0 +1,72 @@
+"""Tree result parser for backend API lookup results.
+
+See backend's app.api.api_v1.routers.lookups.tree_table_to_json
+
+A tree had 0/1/many nodes.
+A node typically looks like this:
+
+    {
```
joel-wright, line 8: I don't think we can assume a single root node, so I made the return type for "tree-ish" lookups a list of objects. the "nodes" at the top level are the top level elements (e.g. world bank regions)

joel-wright (commented):
I have made a few comments, and it looks like we still have the join table for categories in the backend database (unless I misunderstood the migration).

backend/app/core/service/loader.py
opyate: Yes, ultimately. 
At the moment, the names on those tables are unique, so we end up with a small subset anyway.
Once we have the migrations for the lookups, perhaps we can look at using lookups then?

loader/app/service/tree_parser.py
opyate: Yeah, it was a silly example. The parser iterates the whole tree, though.

joel-wright (approved):
We still need to answer the question of the backend db category lookup table, but happy to merge to make progress!

PR #511, opened 2022-04-29
Query for new search index.

stefl14:
- Just putting the query up for visibility, but there's a blocker on the data side for fully testing indexing it. 

- A few of the fields are inconsistent with the format assumed on the opensearch backend (i.e. my select * as ___) need to be altered to smoothly interoperate with the data indexer, assuming we don't change anything in opensearch before alpha. But this is the gist of the query. Waiting for other updates to fully test.


loader/app/main.py
```diff
@@ -60,10 +62,17 @@ def main():
     data_dir = get_data_dir()
 
     data = extract(data_dir)
-    policies = transform(data)
+    policies = transform(data)  # noqa: F841
     for db in get_db():
         load(db, policies)
 
+        # # once all data has been loaded into database, upload files to cloud
```
kdutia, line 69: need to ensure these lines are uncommented before merge!

kdutia:
@stefl14 I think the new column names from the query need to be passed to `metadata_columns` in `index_data.get_document_generator` in order for them to be passed to Opensearch?

PR #520, opened 2022-05-04
Create script to transform google sheets excel export to csv

kdutia:
This script takes excel files from the data entry task and single document actions cleanup and maps it to a csv that can be used by the loader. Eventually I imagine this code will be merged into the loader.

You can run `make create_csv` but I've also uploaded the CSV output at `loader/data/loader_data.csv`

Note there's an issue with parsing at the moment which means only ~50% of documents are loaded in, but this is enough to continue with work on getting APIs and the search index up so I've raised issue #524 to fix this.


kdutia:
@opyate I found a bug and raised a new issue #524 so we can get on with more critical stuff right now, so this should be ready for review/merge.

PR #528, opened 2022-05-05
Feature/499-500 passage info and pdf embed slideout

eurolife:
The passage matches and pdf embed slideout panel has been added. It is triggered when clicking on a document title or 'x' matches in document link in each search result.


kdutia (approved):
lgtm 👍

PR #530, opened 2022-05-05
feat-480-add-metadata-to-search-index

stefl14:


search-index/index_data.py
```diff
@@ -222,20 +234,25 @@ def run_cli(
         main_dataset, ids_table, embs, description_embs_dict
     )
 
+    def _convert_to_bool(x):
```
stefl14, line 237: Small helper because OpenSearchIndex needs bools, not the strings parsed from environment variables.

search-index/index_data.py
```diff
@@ -41,38 +41,50 @@ def get_document_generator(
...
-        "action_id",
-        "action_date",
-        "action_name",
-        "action_description",
-        "action_name_and_id",
-        "action_country_code",
-        "action_geography_english_shortname",
-        "action_source_name",
-        "action_type_name",
+        "document_date",
```
stefl14, line 61: Everything prepended by "action" has been removed and replaced with document ids. This list should now contain everything expected according to 480.

search-index/app/load_data.py
```diff
@@ -16,16 +16,115 @@ def get_data_from_navigator_tables(
...
-        geography.*, source.source_id, source.name as action_source_name, action_type.action_type_id, action_type.type_name as action_type_name, action.name as action_name, action.description as action_description, \
-        geography.country_code as action_country_code, geography.english_shortname as action_geography_english_shortname
-        FROM document
-        INNER JOIN action ON (document.action_id = action.action_id)
-        LEFT JOIN language on (document.language_id = language.language_id)
-        LEFT JOIN geography on (action.geography_id = geography.geography_id)
-        LEFT JOIN source on (action.action_source_id = source.source_id)
-        LEFT JOIN action_type on (action.action_type_id = action_type.action_type_id);
-        """
+      SELECT
```
stefl14, line 19: This has everything required by 480 and more (in case we need to update).

kdutia: Just manually checked and seems like languages in the database are broken. Will raise on Slack and Zenhub

search-index/index_data.py
kdutia: A small thing but some code in the backend (see `backend.app.core.config`) which does a similar thing is case-insensitive. Can we make `x.lower() == "true"/"false"` here instead?

pipeline/processor/postprocessor.py
```diff
@@ -720,4 +720,5 @@ def process(self, document: Document, filename: Optional[str]) -> Document:
         doc = document.to_dict()
         new_contents = self._group_list_elements(doc)
         new_contents["filename"] = filename
```
stefl14, line 722: Slight hack here to get around a bit of an antipattern in the post-processor that will results in missing md5 hashes.

search-index/text2embeddings.py
```diff
@@ -359,7 +358,7 @@ def run_cli(
     description_ids_output_path = (
         output_dir / f"description_ids_{model_name}_{curr_time}.csv"
     )
-    description_data_to_encode["prototype_filename_stem"].to_csv(
+    description_data_to_encode["md5_sum"].to_csv(
```
stefl14, line 361: altered this from prototype filename so that the downstream indexing process doesn't break (the loop is now a hash instead of document id).

search-index/index_data.py
```diff
@@ -41,45 +41,54 @@ def get_document_generator(
...
+        "document_description",
     ]
-
-    for document_id, document_df in text_and_ids_data.groupby("document_id"):
-
-        doc_metadata = main_dataset.loc[
-            main_dataset["prototype_filename_stem"] == document_id
-        ]
+    for document_id, document_df in text_and_ids_data.groupby("document_md5_hash"):
+        doc_metadata = main_dataset.loc[main_dataset["md5_sum"] == document_id]
```
stefl14, line 91: To match change in csv created by text2embeddings.

search-index/index_data.py
stefl14: done

search-index/app/load_data.py
```diff
@@ -41,55 +151,3 @@ def ensure_safe(url: str) -> str:
     if "https://" not in url:
         url = url.replace("http://", "https://")
     return url
-
```
stefl14, line 44: Functions removed.

search-index/README.md
```diff
@@ -39,15 +39,20 @@ The following snippets are examples of the structure of different documents in t
...
+          "document_name_and_id" : "National Policy for Disaster Management 1297",
+          "document_country_code" : "KEN",
```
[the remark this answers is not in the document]
kdutia: resolved in #547

kdutia (approved):
✅

PR #535, opened 2022-05-06
Create input csv 

kdutia:


loader/app/loaders/loader_cclw_v2/extract/test_main.py
```diff
@@ -4,7 +4,15 @@
...
+        / ".."
+        / ".."
+        / "data"
+        / "cclw_new_format_20220503.csv"
+    )
 
-    policies_fe = extract(data_dir)
-    assert len(policies_fe) == 87
+    policies_fe = extract(csv_path)
+    assert len(policies_fe) == 1371
```
opyate, line 18: ooh, so many rows now...

opyate (approved):

loader/app/loaders/loader_cclw_v2/extract/test_main.py
kdutia: but the wrong number because i found another error in the input data 😂

PR #539, opened 2022-05-06
some changes to get frontend to work on eb

opyate:


infra/cpr/backend/main_with_frontend.py
```diff
@@ -92,29 +96,52 @@ def __init__(
...
+      OPENSEARCH_PASSWORD: %s
+      OPENSEARCH_URL: %s
 
   frontend:
     image: %s
     mem_limit: 256m
-    command: npm run prod
+    command: npm run start
+    environment:
+      PORT: 3000
```
opyate, line 118: We probably need `NEXT_PUBLIC_API_URL` here.

nginx/nginx.conf
```diff
@@ -20,4 +20,13 @@ server {
...
 	}
+
+	location /health {
+	    proxy_pass http://backend:8888/health;
+	}
+
+	location ~ ^/test/(.*)$ {
+	    resolver 8.8.8.8;
+	    proxy_pass https://httpbin.org/$1$is_args$args;
+	}
```
opyate, line 31: This was to test to test nginx on its own. I was wondering why things weren't working, but looked like it must. So I tried this locally:

image


Then I tried it on the EB public URL, and it didn't work. Then after I fixed the health check, the above, and frontend+API endpoints started working. Ruddy AWS!

joel-wright (approved):
Let's get this merged

PR #540, opened 2022-05-08
Feature 536 - Enable Sendgrid

joel-wright:
This PR enables sendgrid with templated emails for the following:
- New account activation
- Password reset request
- Password updated notification


backend/app/api/api_v1/routers/admin.py
```diff
@@ -90,7 +93,7 @@ async def user_edit(
     """Updates existing user"""
     updated_user = edit_user(db, user_id, user)
 
-    send_email(EmailType.account_changed, updated_user)
+    # send_email(EmailType.account_changed, updated_user)
```
kdutia, line 96: What's the reason for not sending an email when a user's account details are edited? Is this just a thing we'll activate later?

kdutia (approved):
lgtm!

backend/app/api/api_v1/routers/admin.py
joel-wright: yeah, basically we don't have a template yet :)

PR #543, opened 2022-05-09
Add switch to disable API docs

kdutia:
Added a `ENABLE_API_DOCS` environment variable to enable/disable API docs at `/api/docs` and openAPI json at `/api`.

For reasons I don't understand you need to completely rebuild the backend container for `backend/app/main.py` to see a change in this environment variable, even though python sees the change (as tested by `docker-compose run backend python -c "import os; print(os.environ['ENABLE_API_DOCS'])"`).


joel-wright (approved):
lgtm

PR #545, opened 2022-05-09
Add GET document endpoint (multiple queries)

joel-wright:
First pass version with multiple queries


backend/app/db/crud/document.py
```diff
@@ -47,3 +48,14 @@ def is_document_exists(
             )
         )
     ).scalar()
+
+
+def get_document(db: Session, document_id: int) -> Document:
+    document = db.query(Document).filter(Document.id == document_id).first()
```
opyate, line 54: Also filter by `is_valid`?

joel-wright: Ah okay, I'll add that - I thought that invalid documents weren't pushed into the api db

opyate: Ah, maybe you're right, and there's nothing to do.

opyate (approved):

PR #546, opened 2022-05-10
Feature/498 cover page

eurolife:
Document cover page added using sample data. 
Some other fixes/features include:
- In search results, click on Litigation tab and a 'coming soon' message will display
- After selecting some filters, sorting, etc, and then leaving the page, then coming back to the page, these selections will still be shown in the UI (previously they were lost when navigating away).


opyate (approved):

kdutia:
@eurolife not sure whether this is to do with the addition of the litigation tab but the download results button seems to be out of place? 

Although, we mentioned hiding this button until we have the API to download results. Not sure whether we decided on that or showing a 'feature coming soon' message. 

image

kdutia:
Also, I'd expect the document title here to link to the document cover page? Because that functionality is otherwise hidden in an overflow menu, someone might never realise they can get to the document cover page from the search results.
image

kdutia:
On the document cover page, do these 'link to document' links imply that we have a copy of the document for each event? I don't think we have that in the data at the moment.

image

eurolife:
> @eurolife not sure whether this is to do with the addition of the litigation tab but the download results button seems to be out of place?
> 
> Although, we mentioned hiding this button until we have the API to download results. Not sure whether we decided on that or showing a 'feature coming soon' message.
> 
> image

@kdutia I'm not able to reproduce this one - maybe you need to clear cache?

eurolife:
> On the document cover page, do these 'link to document' links imply that we have a copy of the document for each event? I don't think we have that in the data at the moment.
> 
> image

@kdutia I was just picking that up from the design - I'll have to ask Ingemar about it.

eurolife:
> Also, I'd expect the document title here to link to the document cover page? Because that functionality is otherwise hidden in an overflow menu, someone might never realise they can get to the document cover page from the search results.
> 
> image
> 
> The 'View Document Details' link in the overflow menu doesn't work at the moment either, but I'm not sure if you had that planned for a later PR.
> 
> image

@kdutia All good points here. I will take care of these in my next PR.

PR #547, opened 2022-05-10
Feat 494 event dates

opyate:


backend/app/db/models.py
```diff
@@ -343,7 +343,7 @@ class Event(Base):  # noqa: D101
     )
     name = sa.Column(sa.Text, nullable=False)
     description = sa.Column(sa.Text, nullable=False)
-    created_ts = sa.Column(sa.DateTime(timezone=True), server_default=func.now())
```
joel-wright, line 346: I thought we also decided to add an explicit year (or even date) field to the `Document` model too?

kdutia (approved):
lgtm!

joel-wright (approved):
LGTM, just the one question about the explicit year/date we wanted to add to the `Document` table. That can be done in a follow-up if we still decide we need it after a quick chat.

backend/app/db/models.py
opyate: Yes, after merging your currently open PR?

opyate: Yes, after merging your currently open PR?

PR #554, opened 2022-05-11
doc associations

opyate:
Backend `get_current_user` to only decode JWT and not fetch user from DB as this starved the DB/pool of connections.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -56,8 +56,6 @@ SECRET_KEY=super_secret
 # machine user account for loader
 MACHINE_USER_LOADER_EMAIL=loader
 MACHINE_USER_LOADER_PASSWORD=password
-# machine user token has no expiry
-MACHINE_USER_LOADER_JWT=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJsb2FkZXIiLCJwZXJtaXNzaW9ucyI6ImFkbWluIn0.GRymxg-mbw2oO2zEBQm8p8eS7MqUSvBolljWgTufEcI
 
 PASSWORD_RESET_TOKEN_EXPIRY_MINUTES=30
 
--- a/backend/app/core/auth.py
+++ b/backend/app/core/auth.py
@@ -8,16 +8,18 @@ from app.core import security
 from app.db import session
 from app.db.crud.user import get_user_by_email
 from app.db.models import User
+from app.db.schemas.user import JWTUser
 
 
-async def get_current_user(
-    db=Depends(session.get_db), token: str = Depends(security.oauth2_scheme)
-):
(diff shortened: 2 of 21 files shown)


.env.example
```diff
@@ -56,8 +56,6 @@ SECRET_KEY=super_secret
 # machine user account for loader
 MACHINE_USER_LOADER_EMAIL=loader
 MACHINE_USER_LOADER_PASSWORD=password
-# machine user token has no expiry
-MACHINE_USER_LOADER_JWT=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJsb2FkZXIiLCJwZXJtaXNzaW9ucyI6ImFkbWluIn0.GRymxg-mbw2oO2zEBQm8p8eS7MqUSvBolljWgTufEcI
```
opyate, line 60: Machine user now just logs in before POSTing to backend.

backend/app/core/auth.py
```diff
@@ -8,16 +8,18 @@
 from app.db import session
 from app.db.crud.user import get_user_by_email
 from app.db.models import User
+from app.db.schemas.user import JWTUser
```
opyate, line 11: The main change here is that there are now 2 ways to inject a user dependency in the API endpoints:
- get_current_active_user 
- get_current_active_db_user
(and their superuser analogues)

The former is quick (only a JWT decode) and the latter is heavy (fetches user from DB).
This sped up the doc upload quite a bit, as it doesn't need the DB user, but the request is kept open for a long time as Boto uploads PDFs to S3. This starved the DB pool with the original dependency.

loader/app/db/schema.py
```diff
@@ -0,0 +1,14 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class AssociationSchema:
+    """A remote association, to be posted to backend.
+
+    TODO move to common schemas.
```
opyate, line 8: Maybe one day we can move all schemas to common. But now right now :)

loader/app/loaders/loader_cclw_v2/load/main.py
```diff
@@ -62,6 +62,13 @@ async def load(ctx: Context, policies: PolicyLookup):
     logger.info(f"Done, imported {doc_count} docs from {len(policies.items())} actions")
 
 
+def warmup_local_caches():
```
opyate, line 65: As some of the loader POSTs/uploads are now called from futures, best to warm up the cache outside of a possible stampeding herd.

loader/app/db/schema.py
opyate: Also, then change this from a dataclass to a pydantic model.

joel-wright (approved):
LGTM

PR #555, opened 2022-05-12
Feature/501 landing page

eurolife:
The landing page has been updated as per the latest Figma design.


kdutia (approved):
looks good! I didn't realise there was stuff to scroll to because the background really nicely fits to the height of my browser window (not sure whether this was intentional). 

Maybe a nice future addition could be making the background a little bit shorter, or some other way of indicating that there's content to scroll to below the search box and suggested searches.

PR #562, opened 2022-05-12
Loader fixes

kdutia:
Found a couple of bugs where the some metadata fields were wrongly split from the CSV, and posting documents from the loader DB to the API DB was failing.

### DIFF
--- a/loader/app/poster/post.py
+++ b/loader/app/poster/post.py
@@ -100,7 +100,7 @@ def post_doc(db: Session, doc: Document) -> dict:
             "source_id": doc.source_id,
             "geography_id": doc.geography_id,  # this is from backend API lookup, so will exist remotely.
             "category_id": doc.category_id,  # this is from backend API lookup, so will exist remotely.
-            "publication_ts": doc.publication_ts,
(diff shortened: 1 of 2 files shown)


joel-wright (approved):
LGTM

loader/app/poster/post.py
```diff
@@ -100,7 +100,7 @@ def post_doc(db: Session, doc: Document) -> dict:
             "source_id": doc.source_id,
             "geography_id": doc.geography_id,  # this is from backend API lookup, so will exist remotely.
             "category_id": doc.category_id,  # this is from backend API lookup, so will exist remotely.
-            "publication_ts": doc.publication_ts,
+            "publication_ts": doc.publication_ts.isoformat(),
```
opyate, line 103: This is fixed in https://github.com/climatepolicyradar/navigator/pull/554 BTW :+1:

joel-wright: the new lines looked like they matched so I thought there would probably be no conflicts

joel-wright: :grin:

opyate (approved):

PR #563, opened 2022-05-13
Feature/add login env var

eurolife:
Env variable added for login api.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -48,7 +48,7 @@ AWS_REGION=eu-west-2
 
 # Frontend
 NEXT_PUBLIC_API_URL=http://localhost:8000/api/v1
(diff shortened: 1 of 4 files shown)


.env.example
```diff
@@ -48,6 +48,7 @@ AWS_REGION=eu-west-2
 
 # Frontend
 NEXT_PUBLIC_API_URL=http://localhost:8000/api/v1
+NEXT_PUBLIC_LOGIN_API_URL=http://localhost:8000/api/v1
```
opyate, line 51:
```suggestion
NEXT_PUBLIC_LOGIN_API_URL=http://localhost:8000/api/tokens
```

eurolife: > 

oops, fixed!

opyate (approved):

PR #564, opened 2022-05-13
Update /searches API

kdutia:
- remove all mentions of actions
- add new metadata
- add URLs which are mapped from s3 to our CDN

There's a bug which means they're only available through `document_instrument_name` and `document_instrument_parent` is always blank in Opensearch at the moment, but that's reflected in the keyword filtering args available through the API request structure. It also means that the instrument lookup won't match the range of possible instrument values for now. I'll open another PR to fix this.

Sample request and response:
image


joel-wright (approved):
Updates opensearch to match database, so approving with existing TODOs

PR #566, opened 2022-05-13
Cors issue

opyate:


joel-wright:
Seems to not be required. We can re-open if necessary.

joel-wright:
Seems to not be required. We can re-open if necessary.

PR #608, opened 2022-05-23
Non-pdf UI changes

eurolife:
This adds logic which checks for a document 'content_type' that is 'application/pdf'. 

If it is a PDF, a PDF download icon/button will be visible on the document cover page.

If the document is a PDF and has passage matches, the 'X matches in  document' link will be visible in search results and clicking it will open the passage matches slideout panel, otherwise this link won't be displayed.

The document title will now always go to the document cover page when clicked.

Document cover page clickable icon has been removed from search results.

This PR depends on the search and document APIs to return a top level 'content_type' value.


joel-wright (approved):
LGTM

PR #610, opened 2022-05-23
Fix 605: Create document entries for all content types & languages

joel-wright:
~TODO: testing, but could do with a sanity check that this approach will work for Opensearch~


kdutia:
lgtm - tested on a small dataset and rewrote the opensearch indexer to propagate these changes to the search index. Will tidy up and open a PR soon

kdutia (approved):

PR #620, opened 2022-05-25
Fix - add missing docs

kdutia:
* 275 docs without URLs have been added
* `make create_csv` now logs the number of docs without titles (currently 54) from the data entry exercise


joel-wright (approved):
LGTM

PR #638, opened 2022-05-27
text2embeddings runs with suitable efficiency

kdutia:
When the number of md5sums increased by an order of magnitude, part of the text2embeddings process became painfully slow.
We were also getting a separate progress bar printed for each batch, as well as the main one, so I've disabled that.

### DIFF
--- a/search-index/app/ml.py
+++ b/search-index/app/ml.py
@@ -59,7 +59,9 @@ class SBERTEncoder(SentenceEncoder):
         Returns:
             np.ndarray
         """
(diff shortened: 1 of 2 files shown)


stefl14 (approved):
Looks perfectly sensible.

joel-wright (approved):
LGTM, let's really take a good look at this when we come to defining a nice repeatable pipeline. I feel like we're going to get 150,000 documents dropped on us at some point :joy:

PR #642, opened 2022-05-30
Fix/622 sort structure lookups

eurolife:
This will sort filters alphabetically and display the instruments in a better structure in both filters and on the document cover page.


joel-wright (approved):
LGTM

PR #652, opened 2022-06-01
Fix/search results fixes

eurolife:
This PR includes the following improvements/fixes:

- pagination no longer displayed when no results returned
- annotation highlighting on PDFs hidden until a solution is found to make it work correctly
- UI responded with a delay on choosing filters and opening and closing passage matches panel, also search sent another api request when returning to search page from doc cover page which was unneeded. All of this fixed now.
- Activate account page now has text and link to sign in for users that already have an account
- User can now update their password from their account section when logged in
- User now gets confirmation when they update their account details


joel-wright (approved):
LGTM

PR #661, opened 2022-06-06
Fix/stop unneeded search api calls

eurolife:
This includes some improvements including:

- When visiting the search page, search api calls are only made when coming from the landing page or entering search terms. When returning from document cover page or other non-landing page, the previous search is retained and displayed without a new api call.
- Passage matches links still toggle the slideout panel unless switching between documents, then the panel updates with the new set of passage matches
- Full page PDF view expanded more in height, error that popped up when no passage selected is resolved.


joel-wright (approved):
LGTM

PR #662, opened 2022-06-06
Bug/pdf jump to page

eurolife:
This fixes a bug where the first passage match would not jump to its page number when clicking to view it in the pdf.


joel-wright (approved):
LGTM

PR #684, opened 2022-06-20
encode password on signin

eurolife:
Encode password on user signin


joel-wright:
This looks good to me, I'll deploy to dev so we can test the fix :)

joel-wright (approved):
LGTM

PR #687, opened 2022-06-22
Feature/fro 26 password reveal

PatFawbertMills:
FRO-26 Adding a password reveal button to the sign in form (and potentially anywhere that uses a password input)


frontend/src/components/form-inputs/PasswordInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { FieldErrors } from 'react-hook-form';
+import FormFieldError from '../blocks/Error';
+import { EyeIcon } from "../../components/svg/Icons";
+
+interface InputProps {
+  label?: string;
+  required?: boolean;
+  errors: FieldErrors;
+  name: string;
+  type?: string;
```
eurolife, line 11: Probably don't need the 'type' prop since it will always be a password field in this case.

frontend/src/components/form-inputs/PasswordInput.tsx
```diff
@@ -0,0 +1,66 @@
...
+        {label}
+        {required && label ? (
+          <strong className="text-red-500"> *</strong>
+        ) : null}
+      </label>
+      <div className="relative mt-1">
+        <input
+          type={passwordReveal ? 'text' : type}
+          placeholder={placeholder}
+          accept={accept}
```
eurolife, line 49: Not sure we need the 'accept' prop in this case.

eurolife (changes requested):
Please also update the password/password confirm fields on the auth/activate-account and auth/reset-password pages. You might also want to update it on the account/change-password page as well even though it's currently not being used.

eurolife (approved):

PatFawbertMills:
@joel-wright don't merge until I've got my own account to test the pages

joel-wright (approved):
LGTM

PR #692, opened 2022-07-04
Feature/fro20 improve date range selection

PatFawbertMills:
FRO-20

Previous date picker was difficult to use so has been adapted to allow a 'last 5 years', 'last 10 years' and custom date range input. Additional options can be added trivially when required.

There is also some validation and feedback around the custom dates to prevent user error.


joel-wright (approved):
LGTM

PR #693, opened 2022-07-04
Add required fields to opensearch queries

joel-wright:
Allows us to fix searching when document name is empty in opensearch database.

### DIFF
--- a/backend/app/core/search.py
+++ b/backend/app/core/search.py
@@ -1,7 +1,7 @@
 import time
 import os
 from dataclasses import dataclass
-from typing import Any, Dict, List, Mapping, Optional, Tuple
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple
 
 from opensearchpy import OpenSearch
 from sentence_transformers import SentenceTransformer
@@ -460,7 +460,7 @@ class QueryBuilder:
         else:
             raise RuntimeError("Unknown sort ordering field: {field}")
 
-    def with_required_fields(self, required_fields: List[str] = ["document_name"]):
+    def with_required_fields(self, required_fields: Sequence[str]):
         """Ensure that required fields are present in opensearch responses."""
         must_clause = [
             {"exists": {"field": field_name}} for field_name in required_fields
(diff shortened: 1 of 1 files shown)


backend/app/core/search.py
```diff
@@ -456,6 +460,13 @@ def order_by(self, field: SortField, order: SortOrder):
         else:
             raise RuntimeError("Unknown sort ordering field: {field}")
 
+    def with_required_fields(self, required_fields: List[str] = ["document_name"]):
```
kdutia, line 463: do we want to get rid of the default value of `required_fields` here as it's hardcoded above? (i know it was in my example but it was lazy coding because it followed a bunch of other kwargs with defaults 😅)

kdutia:
This might be a good opportunity to ensure that all fields in `OpenSearchResponseMatchBase` are present?

``` py
class OpenSearchResponseMatchBase(BaseModel):
    """Describes matches returned by an OpenSearch query"""

    document_name: str
    document_country_code: str
    document_description: str
    document_source_name: str
    document_id: int
    document_date: str
    document_name_and_id: str
    document_country_english_shortname: str
    document_type: str
    document_source_url: str
    document_url: str
```

backend/app/core/search.py
joel-wright: ugh, I meant to change that! using mutables as kwargs is really bad practice & I forgot to remove it! thanks!

joel-wright:
> This might be a good opportunity to ensure that some or all fields in `OpenSearchResponseMatchBase` are present?
> 
> ```python
> class OpenSearchResponseMatchBase(BaseModel):
>     """Describes matches returned by an OpenSearch query"""
> 
>     document_name: str
>     document_country_code: str
>     document_description: str
>     document_source_name: str
>     document_id: int
>     document_date: str
>     document_name_and_id: str
>     document_country_english_shortname: str
>     document_type: str
>     document_source_url: str
>     document_url: str
> ```

I think we can follow up on this later; the PR as presented gets us past a live issue so it would be good to get the fix in testing while we decide exactly which fields we need to enforce... does that seem reasonable?

kdutia (approved):

PR #702, opened 2022-07-11
Set default AWS region

joel-wright:
Avoids crash on startup when region is not defined.


joel-wright:
I'll take responsibility for this small change

PR #711, opened 2022-07-19
Feature/fro 63 document slide out

PatFawbertMills:
Improve the UX and functionality of the document slide out and passage selection:
- PDF preview is displayed alongside passages 
- PDF scrolls to the relevant location in the document when a passage is selected
- On mobile devices the PDF and passages are split
- PDF is not longer re-rendered each interaction
- New reusable usePDFPreview hook 🪝


diversemix (approved):
Nice

PR #712, opened 2022-07-20
Feature/fro 66 sectors multi select

PatFawbertMills:
Update to the sectors filters 
- Allow selection of multiple filters
- Intelligent labels for select box depending on selection - default to 'All'
- Strongly typed TSector
- Allow customising of select default copy


diversemix (approved):
lgtm ... my only comment would be to check what happens / if its possible for `actions` being null

diversemix (approved):
Great :+1:

PR #713, opened 2022-07-21
fix backend search tests

kdutia:
Created a new Opensearch dump and changed the keyword used for search for broken searches to 'climate'. Even with the new larger dump 'disaster' wasn't quite returning enough results for a bunch of the tests to pass.

closes DSCI-50


diversemix (approved):
lgtm - apologies I did look at this on Friday but forgot to approve :cry:

PR #715, opened 2022-07-25
Convert delay to a constant and speed it up

PatFawbertMills:
Waiting time of 500ms is not necessary anymore due to the change that renders the PDF before the passage selection is interacted with and makes the app feel sluggish.

This is also now controlled by a constant.


diversemix (approved):
Great change, might be worth testing on mobile and another browser too?

PR #717, opened 2022-07-26
remove test_pipeline and test_search_index from github actions

kdutia:
We moved the pipeline and search indexer to separate repos but i forgot to delete the github actions which trigger the make commands to build and test these in this repo.

At the moment these [complete in a few seconds](https://github.com/climatepolicyradar/navigator/actions/runs/2738658864) and sometimes confuse github into believing they never happened, causing it to wait infinitely for them to finish


PatFawbertMills (approved):
Thanks for removing!

PR #722, opened 2022-08-02
Fix BAK-1051: cleanup duplicate metadata table entries & associated bug

joel-wright:
This PR does the following:
- closes off a bug that would create duplicate metadata (by limiting document creation to existing metadata fields)
- provides a script that cleans up duplicate keywords & instruments in an existing database
- completes an existing TODO so that documents & metadata links are created in a transaction

We should discuss how we should test a script that should probably be removed soon...

### DIFF
--- a/backend/app/api/api_v1/routers/documents.py
+++ b/backend/app/api/api_v1/routers/documents.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from datetime import datetime
+from http.client import INTERNAL_SERVER_ERROR, UNPROCESSABLE_ENTITY
 from pathlib import Path
 from typing import cast
 
@@ -19,7 +20,7 @@ from app.core.auth import (
     get_current_active_db_superuser,
 )
 from app.core.aws import get_s3_client
-from app.core.service.loader import persist_document_and_metadata
+from app.core.service.loader import UnknownMetadataError, persist_document_and_metadata
 from app.core.util import CONTENT_TYPE_MAP, content_type_from_path, s3_to_cdn_url
 from app.db.models import (
     Association,
@@ -68,12 +69,12 @@ from app.db.schemas.metadata import (
 )
 from app.db.session import get_db
 
-logger = logging.getLogger(__file__)
+_LOGGER = logging.getLogger(__file__)
 
-documents_router = r = APIRouter()
+documents_router = APIRouter()
 
 
-@r.get(
+@documents_router.get(
     "/documents/{document_id}",
     response_model=DocumentDetailResponse,
     response_model_exclude_none=True,
@@ -276,7 +277,7 @@ async def get_document_detail(
     )
 
 
-@r.post("/documents", response_model=DocumentInDB)
+@documents_router.post("/documents", response_model=DocumentInDB)
 async def post_document(
     request: Request,
     document_with_metadata: DocumentCreateWithMetadata,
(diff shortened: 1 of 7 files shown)


backend/app/api/api_v1/routers/documents.py
```diff
@@ -320,24 +327,24 @@ def document_upload(
         )
     except Exception:
         raise HTTPException(
-            500,
+            INTERNAL_SERVER_ERROR,
```
diversemix, line 330: Is there not an import for the 415 we are returning on line 319?

backend/app/core/service/loader.py
```diff
@@ -51,17 +107,17 @@ def persist_document_and_metadata(
 
 
 def write_metadata(
-    db: Session,
+    db,
```
diversemix, line 110: Why was the type removed?

backend/app/core/service/loader.py
```diff
@@ -73,147 +129,104 @@ def write_metadata(
...
-                .filter(Sector.name == meta.name)
-                .filter(Sector.source_id == db_document.source_id)
-            ).first()
-            meta_id = existing_sector.id  # type: ignore
-
-        db_bridge = DocumentSector(
-            sector_id=meta_id,
+    for sector in document_with_metadata.sectors:
+        # A sector should already exist, so fail if we cannot find it
+        existing_sector = (
```
diversemix, line 139: Being that we only ever get the id ... can we not:
```
           sector_id = db.query(Sector.id)
            .filter(Sector.name == sector.name)
            .filter(Sector.source_id == db_document.source_id).scalar()
```

backend/app/core/service/loader.py
```diff
@@ -73,147 +129,104 @@ def write_metadata(
...
-                .filter(Instrument.name == meta.name)
-                .filter(Instrument.source_id == db_document.source_id)
-            ).first()
-            meta_id = existing_instrument.id  # type: ignore
-
-        db_bridge = DocumentInstrument(
-            instrument_id=meta_id,
+    for instrument in document_with_metadata.instruments:
+        # An instrument should already exist, so fail if we cannot find it
+        existing_instrument = (
```
diversemix, line 157: Ditto

backend/scripts/cleanup_lookups.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+LookupID = Mapping[str, int]
+
+
+def get_model_values(db: SessionLocal, table: Base) -> LookupID:  # type: ignore
+    id_lookup: Dict[str, int] = {}
+
+    for row in db.query(table).all():
+        should_update_lookup = row.name not in id_lookup or id_lookup[row.name] > int(
```
diversemix, line 23: Does this assume that ids are ascending when added to the table?

backend/app/core/service/loader.py
joel-wright: sorry, meant to put that back - done now

backend/app/core/service/loader.py
joel-wright: done

backend/app/core/service/loader.py
joel-wright: ditto

backend/tests/routes/test_documents.py
```diff
@@ -136,17 +155,99 @@ def test_post_documents(client, superuser_token_headers, test_db):
...
-    assert test_db.query(Hazard).first().name == "some hazard"
-    assert test_db.query(Framework).first().name == "some framework"
-    assert test_db.query(Keyword).first().name == "some keyword"
-    instruments = test_db.query(Instrument).all()
-    assert instruments[0].name == "some instrument"
-    assert instruments[1].name == "another instrument"
     assert test_db.query(DocumentLanguage).first().document_id == 1
 
 
+def test_post_documents_fail(client, superuser_token_headers, test_db):
```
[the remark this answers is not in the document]
joel-wright: we are expecting this to fail because we are attempting to create a document referencing metadata that does not exist...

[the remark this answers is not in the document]
joel-wright: will add to docstring

backend/scripts/cleanup_lookups.py
joel-wright: well, it really doesn't matter, this was just because the original load had ascending IDs and I thought it was aesthetically pleasing to keep the lowest id... I guess I could have just used the first one I found & got exactly the same functionality.

PR #724, opened 2022-08-08
Feature BAK-1128 Adds geography data for countries endpoint

diversemix:
### Work for this PR

- [x] Splits out `models.py` into a module
- [x] Adds the new `geo_statistics` table
- [x] move  backend/alembic/versions/lookups/ backend/app/data_migrations/data
- [x] add populate for geo statistics
- [x] Write tests for init_data


.gitignore
```diff
@@ -179,3 +179,4 @@ dmypy.json
 # Mac OS files
 .DS_Store
 .vim
+models
```
joel-wright, line 182: What's this ignore for?

backend/app/data_migrations/populate_geo_statistics.py
```diff
@@ -0,0 +1,46 @@
+import csv
+from typing import Union
+from app.db.models import (
+    Geography,
+    GeoStatistics
+)
+from app.db.session import SessionLocal
+
+
+def to_float(value: str) -> Union[float, None]:
```
joel-wright, line 10: maybe we should add handling for garbage data here (rather than crashing when we fail to parse a float)?

backend/app/db/models/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+
+These lookups are source-specific:
+- instrument
+- sector
+"""
+
+from .document import *
+from .user import *
+from .source import *
+from .geography import *
```
joel-wright, line 30: technically this file should end with a newline... I thought we had a check for that...

backend/app/initial_data.py
```diff
@@ -49,21 +55,24 @@ def create_loader_machine_user(db) -> None:
     try:
 
         create_user(db,
-            machineuser_email,
-            os.getenv("MACHINE_USER_LOADER_PASSWORD"),
-        )
+                    machineuser_email,
+                    os.getenv("MACHINE_USER_LOADER_PASSWORD"),
+                    )
```
joel-wright, line 60: what horror is this!!? :flushed:

joel-wright (approved):
I'm starting to wonder if we should keep the initial data external to the source code, or maybe just clean up the filenames & use git to track when/how they changed.

Other than that, a couple of minor comments but LGTM.

.gitignore
diversemix: This was because sometimes I run the app locally (not in docker) and link in the models that can be downloaded (see https://github.com/climatepolicyradar/navigator/blob/e28ef9b45a4366231c4510852111a5e6ac3dadac/backend/Dockerfile#L22)

backend/app/data_migrations/populate_geo_statistics.py
diversemix: Ooo good point :+1:

.gitignore
joel-wright: ah ok, so this only ignores `navigator/models`?

diversemix: I've made it specific now ;)

backend/app/data_migrations/populate_geo_statistics.py
diversemix: Also added unit tests

backend/app/initial_data.py
```diff
@@ -22,6 +23,10 @@ def run_data_migrations(db):
     populate_language(db)
     populate_document_type(db)
     populate_geography(db)
+
+    db.commit()  # Geography data is used to geo-stats so commit here
```
joel-wright, line 27: nit: `flush()` would work here & maintain any transaction

backend/tests/test_schema/test_inital_data.py
```diff
@@ -0,0 +1,21 @@
...
+from app.initial_data import populate_initial_data
+
+
+def test_initial_data_populates_tables(engine):
+    helpers = PytestHelpers(engine)
+    helpers.add_alembic()
+
+    with Session(engine) as db:
+        populate_initial_data(db)
+        db.commit()
```
joel-wright, line 12: maybe that `flush()` suggestion was more than a nit...?

joel-wright (approved):
LGTM

backend/app/initial_data.py
diversemix: Nice - good idea

diversemix: Done

backend/tests/test_schema/test_inital_data.py
diversemix: This results in the tests failing:

```
_______________________________________ test_initial_data_populates_tables ________________________________________

engine = Engine(postgresql://navigator:***@backend_db:5432/navigator_test)

    def test_initial_data_populates_tables(engine):
        helpers = PytestHelpers(engine)
        helpers.add_alembic()
    
        with Session(engine) as db:
            populate_initial_data(db)
            db.flush()
    
        with Session(engine) as db:
            geo_count = db.execute("SELECT count(*) FROM geography;").scalar()
            language_count = db.execute("SELECT count(*) FROM language;").scalar()
            geo_stats_count = db.execute("SELECT count(*) FROM geo_statistics;").scalar()
    
>       assert geo_count == 210
E       assert 0 == 210
```

diversemix: sorted

joel-wright (approved):
LGTM

PR #725, opened 2022-08-10
feature/BAK 1129/add geo stats endpoint

diversemix:
Adds the geo_stats endpont

PR #727, opened 2022-08-10
Experiment: router module

joel-wright:
POC for router split up as a module

Demonstrates splitting up routes across multiple files & importing the top level router from the `__init__.py` file of the module.


backend/app/api/api_v1/routers/lookups/__init__.py
```diff
@@ -0,0 +1,15 @@
+# flake8: noqa
```
joel-wright, line 1: disable those pesky unused import warnings

backend/app/api/api_v1/routers/lookups/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+from .lookups1 import (
+    lookup_geographies,
```
joel-wright, line 3: technically it's not necessary to have all these imports, but I had to import something. In the real implementation I would expect that we could import the request/response types?

diversemix:
I think this is really great - its not obvious that the imports are required to "populate" the router - but it works well and keeps things tidy for the time being. Going to be interesting to see how this structure develops - let's keep an eye on it.

Meantime - this change has been incorporated into [this PR](https://github.com/climatepolicyradar/navigator/pull/725) 

I'm happy this is closed now @joel-wright once you've :+1: the above PR :smile:

diversemix (commented):
<3 looks great

joel-wright:
integrated into the geostats pr

PR #728, opened 2022-08-11
BAK-1051: Additional cleanup for typo/messy keywords

joel-wright:
Also includes a desired keyword list for use in initial data loading


backend/scripts/cleanup_lookups/cleanup_lookups.py
```diff
@@ -82,6 +87,62 @@ def cleanup_sector_duplicates(db: SessionLocal) -> None:  # type: ignore
...
+    desired_keywords = set(keyword_map.values())
+    keywords_to_create = desired_keywords - existing_keywords
+
+    name_lookup: Dict[str, int] = {}
+    for k in keywords_to_create:
+        new_keyword = Keyword(name=k, description="Imported by CPR loader")
+        db.add(new_keyword)
+        db.flush()
+        db.refresh(new_keyword)
+        name_lookup[k] = int(new_keyword.id)  # type: ignore
```
diversemix, line 116: My preference would be to add this as a function to the class `Keyword` - in fact what do you think about putting all the db operations for Keywords on the keyword class?

diversemix (approved):
My preference is to keep the db responsibilities together - but seeing this is  a short live script happy to spread it around a little :smile:

backend/scripts/cleanup_lookups/cleanup_lookups.py
joel-wright: After talking about this I think we agreed to push the discussion of these functions to a later date.

PR #730, opened 2022-08-16
Check data does not exist before populating

diversemix:
Checks the tables do not have any rows - if they do - they won't get populated

This has been tested locally with and without existing data.


joel-wright (approved):
LGTM

PR #732, opened 2022-08-17
Replace mypy with pyright

joel-wright:


joel-wright:
Had to ignore some column vs type in sqlalchemy... but I fixed most of it...

diversemix:
On your branch -  I get mypy complaining....
```
mypy .
app/api/api_v1/routers/lookups/geo_stats.py:40: error: Argument "responses" to "get" of "APIRouter" has incompatible type "Mapping[Union[str, int], Mapping[str, Any]]"; expected "Optional[Dict[Union[int, str], Dict[str, Any]]]"
```

diversemix (approved):
lgtm

PR #733, opened 2022-08-18
Move geo_stats endpoint

diversemix:
Preliminary work on BAK-1145

Moves geo_stats from the lookups router to the documents router


diversemix:
No longer needed as we are going to use two endpoints.

PR #734, opened 2022-08-19
BAK-1147 - Add structure for using same "browse_rds" function for summaries page and unfocused search using RDS

diversemix:
Proposed structure of endpoints and code for implementing the browse for RDS and the searching of documents by country.

### DIFF
--- a/backend/.vscode/settings.json
+++ b/backend/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.formatting.provider": "black"
-}
--- a/backend/.vscode/settings.json
+++ b/backend/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.formatting.provider": "black"
+}
--- a/backend/app/api/api_v1/routers/search.py
+++ b/backend/app/api/api_v1/routers/search.py
@@ -53,8 +53,13 @@ def search_documents(
         return browse_rds(search_body)
 
 
-@search_router.post("/searches/country", response_model=CountryCoverPageResponse)
+@search_router.post(
+    "/searches/country/{geography_id}",
+    summary="Gets documents associated with a country by geoography id.",
+    response_model=CountryCoverPageResponse,
+)
 def search_by_country(
+    geography_id: int,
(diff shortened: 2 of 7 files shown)


backend/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.formatting.provider": "black"
```
diversemix, line 2: Just a suggestion to commit this - to make things easier - happy to remove - please vote :+1: / :-1:

joel-wright: I've voted no for the moment, mostly because it seems odd to have IDE specific settings in our repo & I feel like this might conflict... perhaps I'm worrying about nothing...

joel-wright (approved):
LGTM

backend/.vscode/settings.json
PatFawbertMills: I have added some VSCode settings in the FE portion of the repo 👀

joel-wright: how does it interact with local/global settings I already have?

backend/app/api/api_v1/routers/search.py
```diff
@@ -21,23 +32,44 @@
...
+        """When no query string - search using RDS"""
+        return browse_rds(search_body)
+
+
+@search_router.post(
+    "/searches/country/{geography_id}",
+    summary="Gets documents associated with a country by geoography id.",
+    response_model=SearchByCountryResponse,
+)
+def search_by_country(
```
joel-wright, line 67: I've looked at this a couple of times now & keep wondering if this is really a "search"... even the summary string describes it as "get me a summary of this country", rather than "search for country documents".

Would it hurt to update this URL to be something more like `GET /country_summaries/{geography_id}`?

backend/.vscode/settings.json
diversemix: Good point - yeah let me remove this as it may cause issues

joel-wright (approved):
LGTM

PR #735, opened 2022-08-22
Cleanup Document models & lookup all metadata on document creation

joel-wright:


backend/app/db/schemas/document.py
```diff
@@ -12,84 +11,18 @@
...
-
-class _DocumentInDBBase(_DocumentBase):
-    """Properties in the DB.
-
-    Do not use directly.
-    """
+class DocumentOverviewResponse(BaseModel):  # noqa: D101
+    """A document overview returned in browse & related document lists"""
 
     id: int
```
joel-wright, line 25: maybe should be `document_id`

PR #736, opened 2022-08-22
Add country summary endpoint and implement browse_rds

diversemix:
relates to BAK-1148

- Will supersede `document_browse` (which will need removing)
- Browse functionality not implemented (see BAK-1151)
- Country browsing implemented on the summary endpoint 
- Adds tests


joel-wright:
We're going to have some fun merging this with what I've been up to... let's catch up in the morning (sorry :disappointed:)

backend/app/api/api_v1/routers/summaries.py
```diff
@@ -25,8 +26,25 @@ def search_by_country(
...
+
+    for cat in CategoryName:
+        results = browse_rds(db, BrowseArgs(geography_id=geography_id, category=cat))
+        document_counts[cat] = len(results.documents)
+        top_documents[cat] = list(results.documents[:5])
+
+    events = get_events_for_country(db, geography_id)
+
+    # TODO: Add targets
+    targets = []
```
joel-wright, line 43: ugh, targets... something else we need to make a plan for asap!

backend/app/core/browse.py
```diff
@@ -1,12 +1,104 @@
...
+        document_country_english_shortname=row["country_name"],
+        # Stuff we don't use
+        document_source_name="",
+        document_type="",
+        document_source_url="",
+        document_url="",
+        document_content_type="",
+        document_title_match=False,
+        document_description_match=False,
+        document_passage_matches=[],
```
joel-wright, line 43: this looks so much like a `DocumentOverviewResponse`... perhaps we should be trying to unify them...?

joel-wright (approved):
LGTM, just a couple of questions

backend/app/core/browse.py
diversemix: Let's chat - as this is for a response the same as search (as browse needs to return the same structure as search I think) .... /cc @PatFawbertMills

backend/app/api/api_v1/routers/summaries.py
diversemix: lol ... I did mention that to Michal too.

backend/app/core/browse.py
diversemix: Talked - this will be coming with BAK-1151

PR #737, opened 2022-08-22
Bak 1063/improve lookups

THOR300:
***Work in Progress PR*** 

Improve design for lookups - BAK/1063

- Added a new backend endpoint "/config"
- Integrated tests for the endpoint (these successfully pass on the make test_backend command)  
- Integrated response_model for the endpoint 
- Updated the frontend to use the new endpoint and render the response data


THOR300 (commented):
**/.idea should ignore in all folders (globbing)

frontend/src/pages/search/index.tsx
```diff
@@ -62,20 +63,20 @@ const Search = () => {
...
   // get lookups/filters
-  const documentTypesQuery: any = useLookups("document_types");
-  const { data: { data: documentTypes = {} } = {} } = documentTypesQuery;
-
-  const geosQuery: any = useNestedLookups("geographies", "", 2);
-  const { data: { data: { level1: regions = [], level2: countries = [] } = {} } = {} } = geosQuery;
+  const configQuery: any = useConfig("config");
+  const {
+            data: {
+                    document_types: document_types = {},
```
PatFawbertMills, line 69: These collections need to default to empty arrays `[]` as opposed to objects.

frontend/src/hooks/useConfig.ts
```diff
@@ -0,0 +1,65 @@
...
+        geographies,
+        instruments,
+        sectors,
+        regions,
+        countries,
+      }
+      return resp_end;
+    },
+    {
+      onSuccess: (data) => {
```
PatFawbertMills, line 57: Can remove this as it isn't doing anything and we don't need to fire off any side-affects from the query

frontend/src/hooks/useLookups.ts
```diff
@@ -22,7 +22,7 @@ export default function useLookups(path: string, filterProp: string = '') {
     },
     {
       onSuccess: (data) => {
```
PatFawbertMills, line 24: Same as `useConfig` we can remove this

PatFawbertMills (changes requested):
Just need to remove a couple of the query `onSuccess` methods that we don't use

PatFawbertMills (approved):
Looks good nice one 👍

PR #738, opened 2022-08-22
Cleanup Document models & lookup all metadata on document creation

joel-wright:


diversemix (approved):
Lgtm - good first step :raised_hands:

PR #743, opened 2022-08-24
Remove old notebooks

joel-wright:
Old PDF parsing experiment. No longer required, and subsequent development has moved into `navigator-pdf-parser`


diversemix (approved):
lgtm

PR #747, opened 2022-08-25
Prevent political issues by referring to country as geography

diversemix:
Prevent political issues by referring to country as geography.

This is additive for now - we'll remove the other when the front end as changed

### DIFF
--- a/backend/app/api/api_v1/routers/summaries.py
+++ b/backend/app/api/api_v1/routers/summaries.py
@@ -22,6 +22,11 @@ summary_router = APIRouter()
     summary="Gets a summary of the documents associated with a country.",
     response_model=SummaryCountryResponse,
 )
+@summary_router.get(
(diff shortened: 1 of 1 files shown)


backend/app/api/api_v1/routers/summaries.py
```diff
@@ -22,6 +22,11 @@
     summary="Gets a summary of the documents associated with a country.",
```
joel-wright, line 22: are we planning to remove the country endpoint when the FE has been updated?

joel-wright (approved):
lgtm, assuming we can mark the country endpoint as `FIXME: remove`...?

backend/app/api/api_v1/routers/summaries.py
diversemix: yes

PR #750, opened 2022-08-26
bugfix/tsc version issue

diversemix:
This PR does 2 things:
- removes `package-lock.json` from docker ignore file - allowing docker to use it in the container
- uses `npm ci` instead of `npm install` to speed up the build by only using the lock file for deps


THOR300 (approved):

joel-wright (approved):
LGTM

PR #751, opened 2022-08-26
BAK-1152 Add pushing docker images to ECR in CI

diversemix:
wip


joel-wright (approved):
This is really cool, thanks!

PR #757, opened 2022-09-01
Bugfix: Logging crashing on search

diversemix:
Tested locally... one thing to note is that it logs the request body as a string... so escapes the quotes, for example:

```
backend_1                 | {"written_at": "2022-09-01T10:54:35.185Z", ... <snip> ..."search_request": "{\"size\": 0, \"query\": {\"bool\": {\"should\": [{\"bool\": {\"should\": [{\"match\": {\"for_search_document_name\": {\"query\": \"Adaptation strategy\"}}}, {\"match_phrase\": {\"for_search_document_name\": {\"query\": \"Adaptation strategy\", \"boost\": 2}}}], \"boost\": 100}}, .... etc
```

### DIFF
--- a/backend/app/core/search.py
+++ b/backend/app/core/search.py
@@ -1,3 +1,4 @@
+import json
 import time
 import logging
 import os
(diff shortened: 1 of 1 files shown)


joel-wright (approved):
Update looks good, but we have no regression test...

diversemix:
Good point ... let me look at adding one

PR #759, opened 2022-09-01
Remove instruments from filters and component

PatFawbertMills:
Was an issue with old instruments collection vs. new. Wasn't being used so has been removed for now.

Also issue with country link due to switch to new config endpoint.


joel-wright (approved):
Looks reasonable to me...

PR #768, opened 2022-09-02
Ensure sector model is correct for consumption

PatFawbertMills:
Sectors data was being stored unprocessed which was breaking the sectors drop down


joel-wright (approved):
LGTM

PR #772, opened 2022-09-05
feature/bak 1007/opensearch caching

diversemix:
The change implements overriding the parameters as detailed in BAK-1007 so that the first page can be returned more quickly - while the full search is done in the background.

:notebook: This has been tested locally- but no new tests added... Please comment on how / if you think these should be done. I was just going to add some tests for the wrapper without actually doing any searching.


backend/app/core/search_cache_wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+            # override certain config values to return ASAP
+
+            config = opensearch_internal_config
+
+            overrides = {
+                "max_doc_count": 20,
+                "max_passages_per_doc": 5,
+                "n_passages_to_sample_per_shard": 100_000,
+            }
+            config = dataclasses.replace(opensearch_internal_config, **overrides)
```
joel-wright, line 48: This will work fine for now, but it might cause some unexpected behaviour for future API use. Nothing to change now (because we're not making the API public at the moment) but maybe something to note in the code for future developers?

backend/app/core/search_cache_wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+            )
+
+            # In the background do the full query to prime the OpenSearch cache.
+            background_tasks.add_task(
+                SearchCacheWrapper._q,
+                self.os_connection,
+                search_request_body,
+                opensearch_internal_config,
+                preference,
+            )
```
joel-wright, line 60: I assume that the only effect of attempting to get pages beyond the initial fast request before the background task has completed is a potentially slow page load time for page 3?

joel-wright (commented):
Looks nice, FastAPI `BackgroundTask`s are a nice fit & no messing with thread/process pools manually. My only question would be, does FastAPI have any mechanism to identify already running background tasks that perhaps we want to wait for...? (I know, not a simple question!)

backend/app/core/search_cache_wrapper.py
diversemix: We could simply create a new endpoint - one which calls the cache wrapper and one that doesn't - naming them appropriately

backend/app/core/search_cache_wrapper.py
diversemix: yup - I could put a comment in to this effect - but hey - who even gets as far as page 3 in google :smile:

diversemix:
I'll take a look at the background tasks in more detail now and fixup the tests

backend/app/core/search_cache_wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+    ) -> SearchResponseBody:
+        """Caching equivalent of the OpenSearchConnection query function"""
+
+        if background_tasks is not None and self.is_first_page(search_request_body):
+            # override certain config values to return ASAP
+
+            config = opensearch_internal_config
+
+            overrides = {
+                "max_doc_count": 20,
```
diversemix, line 44: Make these env vars

backend/app/core/search_cache_wrapper.py
joel-wright: resolving after adding dev agenda item to discuss

backend/app/core/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import jit_query_wrapper
```
diversemix, line 1: This was necessary to placate pyright:
```
pyright 1.1.267
/home/peter/work/navigator/backend/tests/routes/test_search.py
  /home/peter/work/navigator/backend/tests/routes/test_search.py:136:41 - error: "jit_query_wrapper" is not a known member of module (reportGeneralTypeIssues)
  /home/peter/work/navigator/backend/tests/routes/test_search.py:158:41 - error: "jit_query_wrapper" is not a known member of module (reportGeneralTypeIssues)
2 errors, 0 warnings, 0 informations
```

diversemix, line 1: This was necessary to placate pyright:
```
pyright 1.1.267
/home/peter/work/navigator/backend/tests/routes/test_search.py
  /home/peter/work/navigator/backend/tests/routes/test_search.py:136:41 - error: "jit_query_wrapper" is not a known member of module (reportGeneralTypeIssues)
  /home/peter/work/navigator/backend/tests/routes/test_search.py:158:41 - error: "jit_query_wrapper" is not a known member of module (reportGeneralTypeIssues)
2 errors, 0 warnings, 0 informations
```

kdutia (approved):

backend/app/core/search_cache_wrapper.py
```diff
@@ -0,0 +1,69 @@
...
+                "max_passages_per_doc": 5,
```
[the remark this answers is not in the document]
joel-wright: yeah, I raised this earlier. We need to be returning the same sort of results as the larger query or we're unnecessarily making our top results show fewer results!

joel-wright (commented):
This is looking really good, but we need to make sure that our results aren't unnecessarily restricted for the first 2 pages (our best results!), and maybe take the opportunity to tune the config variables...

backend/app/core/search_cache_wrapper.py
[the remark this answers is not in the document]
diversemix: I've removed this after your comment:
image

backend/app/core/search_cache_wrapper.py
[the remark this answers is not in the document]
diversemix: ok - this is now in config - if you are happy that we make another PR to tweak as per https://github.com/climatepolicyradar/navigator/pull/772#pullrequestreview-1098138636 then please resolve @kdutia

PR #774, opened 2022-09-06
Support language lookup by code & name

joel-wright:


diversemix (approved):
lgtm - as long as all the consumers of `write_metadata` are updated too :+1:

PR #784, opened 2022-09-13
Use plurals in API endpoints as on other endpoints

joel-wright:


joel-wright:
This change aligns the expectations of the `data-ingest` code and `document-relationships` endpoints (and brings the latter in line with the use of plurals elsewhere in the API).

PatFawbertMills (approved):
LGTM

PR #788, opened 2022-09-14
BAK-1155: Update `document` table & associated requests/responses

joel-wright:
- Remove user links
- add slug field
- add import_id field
- update uniqueness constraint to only include source_id & import_id


THOR300 (approved):
LGTM

backend/alembic/versions/0004_modify_document_table_fields.py
```diff
@@ -0,0 +1,72 @@
...
+# revision identifiers, used by Alembic.
+revision = "0004"
+down_revision = "0003"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("document", sa.Column("slug", sa.Text(), nullable=True))
```
diversemix, line 21: Don't we want slugs to be unique?

backend/app/db/models/geography.py
```diff
@@ -25,7 +25,7 @@ class GeoStatistics(Base):
     __tablename__ = "geo_statistics"
 
     id = sa.Column(sa.SmallInteger(), primary_key=True)
-    name = sa.Column(sa.Text(), unique=True)
+    name = sa.Column(sa.Text(), nullable=False, unique=True)
```
diversemix, line 28: I can't see this change reflected in the migrations :thinking:

diversemix (approved):
lgtm - a couple of points/questions

backend/app/db/models/geography.py
joel-wright: I believe this is the default

backend/alembic/versions/0004_modify_document_table_fields.py
joel-wright: done

diversemix (approved):
Changes look good!

joel-wright:
We finally solved the problem of the weird FK constraint. I have tested this a few times, so I think we can proceed to making this available on dev.app

PR #791, opened 2022-09-21
BAK 1186 Migrate code use new doc relationships

diversemix:
This PR removes the old association code and replaces it with the new document relationships.

# DO NOT MERGE UNTIL dev & prod ARE MIGRATED

This code has been tested with the new scripts in the `backend/scripts/testing` folder (see files). A json dump of all the document details on branch `main` was compared to a dump on this branch - using a database snapshot of production. There were no differences encountered as per the diff script in this PR.


backend/app/db/crud/document.py
```diff
@@ -549,12 +543,15 @@ def write_metadata(
...
 ) -> Set[DocumentOverviewResponse]:
-    # Get all associated documents
-    # start with all documents with a reference from this document_id
-    related_documents_to = set(
+    """Gets all the other documents this document is related to."""
+    query_all_relationships_of_document = db.query(
+        DocumentRelationship.relationship_id
+    ).filter(DocumentRelationship.document_id == document_id)
+
+    return set(
```
joel-wright, line 554: as we talked about separately, it would be good at some point in the future to return these as grouped sets of related documents with details about the relationship. That would require a FE update to support, and doesn't really matter at the moment while we only ever have a single relationship for a group of docs anyway! @PatFawbertMills we'll raise a ticket to keep track of this.

backend/app/db/crud/document.py
```diff
@@ -564,96 +561,17 @@ def _get_associated_documents(
             publication_ts=d.publication_ts,
         )
         for _, d, g in (
-            db.query(Association, Document, Geography)
-            .filter(Association.document_id_from == document_id)
-            .join(Document, Association.document_id_to == Document.id)
+            db.query(DocumentRelationship, Document, Geography)
```
joel-wright, line 564: this is so much cleaner!

backend/tests/routes/test_document_relationships.py
```diff
@@ -296,132 +296,3 @@ def test_get_relationship_documents_security(
 
     assert response_get.status_code == 404
     assert response_get.json() == {"detail": "Not Found"}
-
-
-# TODO : Un comment when relationships are going to replace associations
```
joel-wright, line 301: Did this get replaced?

joel-wright (approved):
I love how clean this is, just wondering why the "uncomment this test" was deleted instead?

backend/app/db/crud/document.py
diversemix: Done https://linear.app/climate-policy-radar/issue/BAK-1189/document-relationships-change-behaviour-to-return-all-relationships

backend/tests/routes/test_document_relationships.py
diversemix: Ah yes - I moved this here - but instead now I've  updated the test in `test_documents.py` to validate this when getting the document details. This tests is a bit of a mother-of-all tests and probably needs breaking down but not today :grin:

PR #794, opened 2022-09-26
Fixes arising from new loader exercising endpoints

joel-wright:
I can't recreate the behaviour I see on my test environment (documents & relationships not persisted, although responses work).

### DIFF
--- a/backend/app/data_migrations/data/category_data.json
+++ b/backend/app/data_migrations/data/category_data.json
@@ -6,5 +6,9 @@
   {
     "name": "Policy",
     "description": "Policy"
+  },
+  {
+    "name": "Litigation",
+    "description": "Litigation"
   }
 ]
--- a/backend/app/data_migrations/data/category_data.json
+++ b/backend/app/data_migrations/data/category_data.json
@@ -6,9 +6,5 @@
   {
     "name": "Policy",
     "description": "Policy"
-  },
-  {
-    "name": "Litigation",
-    "description": "Litigation"
   }
 ]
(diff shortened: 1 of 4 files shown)


backend/app/db/crud/document.py
```diff
@@ -163,7 +163,13 @@ def create_document(
...
     ).scalar()
     if existing_geography_id is None:
-        raise UnknownGeographyError(document_create_request.geography)
+        existing_geography_id = (
+            db.query(Geography.id).filter(
+                Geography.value == document_create_request.geography
+            )
+        ).scalar()
+        if existing_geography_id is None:
+            raise UnknownGeographyError(document_create_request.geography)
```
diversemix, line 172: We should perhaps have a function `get_by_value_or_display()` in `crud/geogrpahy.py`

diversemix: Thinking really for testing's sake

diversemix (approved):
lgtm

backend/app/data_migrations/data/category_data.json
```diff
@@ -6,5 +6,9 @@
   {
     "name": "Policy",
     "description": "Policy"
+  },
+  {
+    "name": "Litigation",
```
diversemix, line 11: This should be "Case"

diversemix: In fact I'd prefer we put this in a separate PR

diversemix (changes requested):
Can we not do the Litigation changes in this PR?

backend/app/db/crud/document.py
joel-wright: this has been moved out into a function for geography & language (mainly for the sake of sanity), but sharing a single function was tricky because we need to look at different model attributes & raise different exceptions on failure...

diversemix (approved):
looks good - thanks for moving those other changes out :+1:

PR #801, opened 2022-09-28
Feature: Add update script to add unique IDs derived from external source

joel-wright:


backend/scripts/add_unique_ids/add_unique_ids.py
```diff
@@ -0,0 +1,46 @@
...
+from app.db.models import Document
+from app.db.session import SessionLocal
+
+
+LookupID = Mapping[str, str]
+
+
+def load_id_map() -> Mapping[str, str]:
+    current_folder = pathlib.Path(__file__).parent.absolute()
+    id_map_path = current_folder / "ids_map.json"
```
diversemix, line 17: Where does this file come from? Is it worth adding a README on how it should be used?

backend/scripts/add_unique_ids/extra/map_ids.py
```diff
@@ -0,0 +1,38 @@
...
+            if (name, url) in current_name_urls:
+                print(name)
+            current_name_urls[(name, url)] = r["id"]
+
+    doc_id_to_import = {}
+    with open(cclw_newer_path) as cclw_newer:
+        csv_reader = DictReader(cclw_newer)
+        for r in csv_reader:
+            name = r["Title"].strip()
+            url = r["Documents"].split("|")[0].strip().replace("http://", "https://")
```
diversemix, line 23: why do we do this?

backend/scripts/add_unique_ids/add_unique_ids.py
joel-wright: I really consider this script a one-shot; everything after here should have the IDs and there's no guarantee that what I did this time would continue to work...

backend/scripts/add_unique_ids/extra/map_ids.py
joel-wright: because the CCLW input file is pure bonkers 🤣

diversemix (approved):
lgtm

PR #802, opened 2022-09-29
BAK-1195 Litigation data model

diversemix:
This implements the design as detailed in the [database schema](https://whimsical.com/database-schema-6s3aH2TfckVGpGJWhTgJ4j) and in the [data in the spreadsheets](https://docs.google.com/spreadsheets/d/1B_jNrr_yEaETGZAn89LkqIE3XfOA1t69wVSgLCxZ-9c/edit#gid=1621148161)

Also it was noted that some of the tables would have a max of 32768 rows as they were using `SmallInteger` - this type has been removed from the schema with the migration. This has been tested by checking no results were returned when running:
```
make start
pg_dump -s navigator | grep -i small
```


backend/app/db/models/litiguation.py
```diff
@@ -0,0 +1,187 @@
...
+    case_id = Column(
+        Integer,
+        ForeignKey(Case.id, ondelete="CASCADE"),
+        nullable=False,
+        primary_key=True,
+    )
+
+
+# --------------------------------------------------
+# Laws (Internal & External)
```
diversemix, line 31: @joel-wright I think this is where we should have the discussion about if we have a single type of "ReferencedLaw" object that can be either internal / external with all the fields on we need. At the end of the day we just need to record the referenced law and whether its internal or external is something that can be determined (or change) later.

backend/app/db/models/litiguation.py
```diff
@@ -0,0 +1,187 @@
...
+    lit_document_id = Column(Integer, ForeignKey(LitDocument.id))
+    case_id = Column(Integer, ForeignKey(Case.id))
+
+
+class LitEventDocuments(Base):  # noqa: D101
+    """Database model for the link table between Litigation event and documents."""
+
+    __tablename__ = "lit_event_documents"
+
+    # NB: Ideally we need to check that the event->case is the same as the
```
diversemix, line 183: Remove comment

backend/app/db/models/case.py
```diff
@@ -0,0 +1,100 @@
...
+    """Database model for a Litigation body /  legal entity."""
+
+    __tablename__ = "lit_body"
+
+    id = Column(Integer, primary_key=True)
+    name = Column(Text, unique=True, nullable=False)
+    description = Column(Text)
+
+
+class CaseStatus(enum.Enum):
```
joel-wright, line 25: Subclassing `str` can make these enums more convenient to deal with:

```suggestion
class CaseStatus(str, enum.Enum):
```
e.g. after subclassing `str` `f"{CaseStatus.OPEN}"` would produce `"Open"` rather than `"CaseStatus.OPEN"`, and direct comparison to the values coming out of the CSV files would be possible (e.g. `CaseStatus.OPEN == "Open"` would return `True`)

backend/app/db/models/__init__.py
```diff
@@ -28,3 +28,5 @@
 from .user import *  # noqa F401 F403
 from .source import *  # noqa F401 F403
 from .geography import *  # noqa F401 F403
+from .case import *  # noqa F401 F403
+from .litiguation import *  # noqa F401 F403
```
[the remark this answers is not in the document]
joel-wright: Do we really consider it safe to import `*` rather than the things we want to re-export? `F403` is there for a reason :grin:

joel-wright (approved):
Looks good, just a few comments

backend/app/db/models/litiguation.py
diversemix: That is true - it might incur a higher maintainance when updating than the saving it has when creating - let me scratch that up to a bad idea :smile:

backend/app/db/models/__init__.py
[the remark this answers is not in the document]
diversemix: ok - that is bad of me :smile:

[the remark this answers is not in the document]
diversemix: This should be fixed now

backend/app/db/models/case.py
diversemix: Ooo - going to take a look at this now

diversemix:
This PR has been approved - but its merged into https://github.com/climatepolicyradar/navigator/pull/804 
Reason is that there are some slight adjustments being made to the schema and so as to not generate other migrations and keep things simple it will get merged to main as one single PR /cc @joel-wright

PR #804, opened 2022-09-30
BAK-1193 Script to validate and load data

diversemix:


backend/scripts/cases/data_load.py
```diff
@@ -0,0 +1,24 @@
+import csv
+from typing import Dict
+from config import CSV_PARTIES, CSV_DOCUMENTS, CSV_CASES, CSV_EVENTS
+
+
+def _data_load(filename: str, id_key: str) -> Dict[str, Dict]:
```
joel-wright, line 6:
```suggestion
from config import CSV_PARTIES, CSV_DOCUMENTS, CSV_CASES, CSV_EVENTS


def _data_load(filename: str, id_key: str) -> dict[str, dict]:
```

joel-wright: It is also nice to consider whether we really want `dict[str, dict]` rather than `Mapping[str, Mapping]`... if we never want to modify the structure directly, using the immutable types really does let pyright/mypy identify when we're being bad :smiling_imp:

backend/scripts/cases/verify.py
```diff
@@ -0,0 +1,54 @@
...
+
+def validate_refs(
+    entities: Dict[str, Any], entity_name: str, entity_key: str, primary_set: Set[str]
+):
+    print()
+    print(f"Checking {entity_name}:")
+    referenced_cases = set([e[entity_key] for e in entities.values()])
+    unknown_cases = referenced_cases - primary_set
+    orphaned_cases = primary_set - referenced_cases
+    print(f"  Found {len(referenced_cases)} referenced {entity_key}s.")
```
joel-wright, line 15: Yeah, this looks a lot like what I'm doing. The pattern I've been following is keeping a container to store errors, so that when I get to the end of each validation step I can quickly check if the container is falsey because I've not found any errors to add to it...

joel-wright (commented):
Seems sensible :smiley:

backend/scripts/cases/verify.py
diversemix: Good to know that its following a pattern - I think I'll leave following the reset to when I integrated this into a backend endpoint

diversemix:
Closed see #807

PR #812, opened 2022-10-10
BAK-1216: add endpoints for SEO support

diversemix:
This is a little pre-emptive, but had a slot to fill and expect to use it soon


joel-wright (approved):
lgtm

PR #813, opened 2022-10-10
Update README to reference BrowserStack

joel-wright:


docker-compose.dev.yml
```diff
@@ -8,6 +8,7 @@ services:
       - ./backend:/app/:ro
     ports:
       - "8888:8888"
+    network_mode: host
```
henry-cpr, line 11: Hanging change accidentally added?

henry-cpr (approved):

docker-compose.dev.yml
joel-wright: dammit! thanks

PR #814, opened 2022-10-10
Tidy tests to remove using user token

diversemix:


backend/tests/routes/test_admin.py
```diff
@@ -161,16 +161,19 @@ def test_unauthenticated_routes(client):
...
+    response = client.get(
+        "/api/v1/admin/users/123",
+    )
+    assert response.status_code == 401
     response = client.post(
         "/api/v1/admin/bulk-imports/cclw/law-policy",
-        headers=user_token_headers,
     )
-    assert response.status_code == 404
+    assert response.status_code == 401
```
diversemix, line 176: Arguably this leaks an un-auth now instead of a 404

backend/tests/routes/test_users.py
```diff
@@ -1,13 +1,12 @@
...
 # from unittest.mock import patch
 
-from app.db.models import User
 
-
-def test_authenticated_user_me(client, user_token_headers):
-    response = client.get("/api/v1/users/me", headers=user_token_headers)
-    assert response.status_code == 200
-    assert response.headers.get("Cache-Control") == "no-cache, no-store, private"
+def test_authenticated_user_me(client):
```
diversemix, line 5: Should we just remove this file?

diversemix:
@joel-wright - couple of comments on here - would be good to get your once-over :smile:

backend/tests/routes/test_users.py
joel-wright: I'm not sure if we're going to want these pages for admin users... might be safe to remove them & re-add later if we decide we want them...

backend/tests/routes/test_admin.py
joel-wright: for now, as we talked about, the details of the API aren't secret, so this is probably fine.

backend/tests/routes/test_users.py
diversemix: removed

diversemix:
@joel-wright - also had to update `backend/app/main.py` you might want to double-check that

PR #820, opened 2022-10-14
CI breakdown tasks further to understand where time is going

diversemix:


joel-wright (approved):
lgtm, helpful to know what's going on (and to only run the slower tests if they're likely to pass)

PR #822, opened 2022-10-18
BAK-1212 Add endpoint to update certain document attributes

diversemix:
* Adds `content_type` as a column in the db - continues to infer the type on creation, but returns/updates the value in the db.
* Adds associated migration
* Adds new endpoint PUT `/documents/{id}` that only takes: `md5_sum`, `content_type`, and `source_url`
* Adds tests to check this behaviour
*


backend/app/core/search.py
```diff
@@ -753,6 +753,7 @@ def create_search_response_document(
         document_category=opensearch_match.document_category,
         document_source_url=opensearch_match.document_source_url,
         document_url=s3_to_cdn_url(opensearch_match.document_url),
+        # TODO - we now need to get this from the db
```
diversemix, line 756: Leaving this here for now as a reminder - happy to remove if you think this is not necessary @joel-wright

backend/app/api/api_v1/routers/admin.py
```diff
@@ -228,3 +234,27 @@ async def import_law_policy(
...
+async def update_document(
+    request: Request,
+    id: int,
+    meta_data: DocumentUpdateRequest,
+    db=Depends(get_db),
+    current_user=Depends(get_current_active_superuser),
+):
+    # TODO: As this grows move it out into the crud later.
+    doc_update = update(Document)
+    doc_update = doc_update.values(meta_data.__dict__)
```
joel-wright, line 249: Why not use the builtin model `dict()` method?

```suggestion
    doc_update = doc_update.values(meta_data.dict())
```

joel-wright: This might not scale for the future, because it looks like we'll remove information if we send keys with `null` values... probably ok for now though...

backend/app/core/search.py
joel-wright: something to talk about rather than deciding now I think

joel-wright (approved):
LGTM

joel-wright (approved):
LGTM

PR #816, opened 2022-10-13
FRO-110 Make a start on SSR by doing geo pages

diversemix:
This PR picks up from where @PatFawbertMills left off and completes SSR for the geography pages.

This was tested by saving and formatting the output of:
```
curl http://localhost:3000/geographies/38
```
then searching for the text "There are no documents for Latvia" which was found (see attached)
[geo_38.html.gz](https://github.com/climatepolicyradar/navigator/files/9777433/geo_38.html.gz)

PR #823, opened 2022-10-18
FRO-116: SSR document pages

diversemix:
Applies the same changes that were done for https://github.com/climatepolicyradar/navigator/pull/816

### Tests

- Build - indicated page is now SSR
- Tested locally by using "view page source" - confirmed the data is present
- Tested by using `curl` - confirmed data retrieved from the server


joel-wright (approved):
LGTM

PR #825, opened 2022-10-19
FRO-117: Redirect solution for preserving CCLW SEO

diversemix:
This adds to the configuration a set of redirects that are loaded in from a CSV at app startup

This CSV will take the following format:
```
source, destination, permanent
/test, /, false
```

Where permanent is true/false given:

> Why does Next.js use 307 and 308? Traditionally a 302 was used for a temporary redirect, and a 301 for a permanent redirect, but many browsers changed the request method of the redirect to GET, regardless of the original method. For example, if the browser made a request to POST /v1/users which returned status code 302 with location /v2/users, the subsequent request might be GET /v2/users instead of the expected POST /v2/users. Next.js uses the 307 temporary redirect, and 308 permanent redirect status codes to explicitly preserve the request method used.

*Source*: https://nextjs.org/docs/api-reference/next.config.js/redirects

### DIFF
--- a/frontend/redirects/default.csv
+++ b/frontend/redirects/default.csv
@@ -1,0 +1,1 @@
+/testing-redirects,/,on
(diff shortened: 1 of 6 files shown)


frontend/redirects/default.csv
```diff
@@ -0,0 +1 @@
+/testing-redirects,/,on
```
diversemix, line 1: just a test file for now - suggest leaving this here for testing etc, and any new redirects go into a file with the name reflecting the name of the environment. For example: `cclw-redirects.csv`

joel-wright (approved):
lgtm

PR #826, opened 2022-10-19
Convert document.url to document.cdn_object (stage 1)

joel-wright:


joel-wright:
This may have gone too far... existing URLs will currently be broken... brb

joel-wright:
@diversemix I think this is ready now

PR #828, opened 2022-10-26
enable multi index searching in searches API

kdutia:
Change the behaviour of the search API to work with multiple indices in Opensearch by adding an optional `include_results` request field.

Behaviour:
* if `include_results` is not set by the frontend, the backend will search the default set of opensearch indices, which at this time are `navigator_core` (document names and descriptions) and `navigator_pdfs_non_translated`.
* any fields passed to `include_results` are added to the list of opensearch indices to query


diversemix (approved):
lgtm :parrot:

joel-wright (approved):
lgtm

PR #829, opened 2022-10-26
Feature/fro 120 cookie consent popup

PatFawbertMills:
- Add cookie consent pop up
- Initialise GA and hotjar if consent given
- Save cookie to store preference
- Check for existence of cookie (from either website or app)


diversemix (approved):
lgtm

PR #830, opened 2022-10-26
BAK-1208 update backend document endpoints to use import_id or slug

diversemix:


joel-wright:
Just fixed up a couple of types...

joel-wright (approved):
LGTM

PR #831, opened 2022-10-26
BAK-1218: Implement document slug spec

joel-wright:


backend/app/db/crud/document.py
```diff
@@ -155,11 +157,27 @@ def __init__(self, category: str) -> None:
...
+) -> str:
+    return "_".join(
+        [
+            slugify(f"{geography.display_value}", separator="-"),
+            f"{publication_ts.year}",
+            slugify(document_name, separator="-"),
+        ]
+    )[
+        :1024
+    ]  # just limit the slug to a reasonable length
```
diversemix, line 173: We need some sort of mechanism to check/retry for non-uniquess

diversemix (approved):
Looks like a good first go - let's see what happens :smile:

backend/app/db/crud/document.py
joel-wright: you're right, and we should track that, but for now I've used the unique components of the import id to cheat a bit

PR #834, opened 2022-10-27
BAK-1208 create document entries on bulk import

joel-wright:


THOR300 (approved):
Please can you update the tests the new object type for the response post our chat over huddle.!

PR #839, opened 2022-10-31
Make allowed origins for CORS explicit

diversemix:


backend/app/initial_data.py
```diff
@@ -117,7 +118,10 @@ def wait_for_app():
 
 if __name__ == "__main__":
     print("Creating initial data...")
-    wait_for_app()
+    skip_wait = len(argv) > 1
+
+    if not skip_wait:
+        wait_for_app()
```
joel-wright, line 124: what's this bit all about?

backend/app/main.py
```diff
@@ -74,10 +74,16 @@
...
+        "http://localhost:3000",
+        "https://app.climatepolicyradar.org",
+        "https://cclw.climatepolicyradar.org",
+        "https://app.dev.climatepolicyradar.org",
+        "https://cclw.dev.climatepolicyradar.org",
+    ],
     allow_credentials=True,
     allow_methods=["*"],
     allow_headers=["*"],
-    allow_origin_regex=r"https:\/\/.*\.climatepolicyradar\.org|http:\/\/localhost:3000",
```
joel-wright, line 80: this is a real shame :(

joel-wright (approved):
lgtm

backend/app/initial_data.py
diversemix: When running from the cli and against a remote host you've tunnelled too - you don't want to wait for the app as it will be the wrong config

joel-wright: I meant more that this change looks like "chuck any old garbage on the command line to skip checks"...

joel-wright: I also can't immediately see how it has an effect

diversemix: @joel-wright Also writing this up as part of `navigator-scripts` now

PR #842, opened 2022-10-31
Parser requires document slug

joel-wright:


backend/tests/routes/test_admin.py
```diff
@@ -477,7 +477,7 @@ def test_bulk_import_cclw_law_policy_preexisting_db_objects(
     assert response_json["document_skipped_ids"] == [existing_doc_import_id]
 
     mock_write_s3.assert_called_once()
-    assert len(mock_write_s3.mock_calls[0].kwargs["documents"]) == 2
+    assert len(mock_write_s3.mock_calls[0].kwargs["documents"]) == 1
```
diversemix, line 480: Did we also want to check the filename now has periods `.` ?

diversemix (approved):
lgtm

backend/tests/routes/test_admin.py
joel-wright: after speaking we decided finer validation can wait

PR #843, opened 2022-11-01
Need to escape '.' in import id matcher :(

joel-wright:


diversemix (approved):
Ah that was my bad - good fix!

PR #844, opened 2022-11-01
Use slugs not document_id(s)

diversemix:


frontend/public/testdata/searches.json
```diff
@@ -9,7 +9,7 @@
...
       "document_description_match": true,
       "document_geography_english_shortname": "Kenya",
-      "document_id": 1299,
```
joel-wright, line 12: so, in the search results we will still have `document_id` (but it will now be an `import_id`), and the slug field will be `document_slug`

joel-wright (commented):
This seems entirely reasonable, but I have one general comment to make. The results coming back from search will:

- continue to include `document_id` (now the `import_id`)
- include a `document_slug` field (not the bare word `slug` here)

frontend/public/testdata/searches.json
diversemix: Thanks - let me fix that

joel-wright (approved):
lgtm

PR #850, opened 2022-11-01
Change bulk import to put create documents in background

diversemix:


backend/app/api/api_v1/schemas/document.py
```diff
@@ -173,7 +173,7 @@ class BulkImportValidatedResult(BaseModel):
     document_count: int
     document_added_count: int
     document_skipped_count: int
-    document_skipped_ids: list[str]
+    document_skipped_ids: set[str]
```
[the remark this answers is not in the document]
joel-wright: looks like the pydantic `.json()` method handles it, must've been when I was doing it manually

joel-wright (approved):
lgtm

backend/app/api/api_v1/routers/admin.py
```diff
@@ -198,57 +200,32 @@ async def import_law_policy(
...
+        import_ids_to_create = []
         # TODO: Check for document existence?
         for validation_result in extract_documents(
             csv_reader=csv_reader, valid_metadata=valid_metadata
         ):
             if validation_result.errors:
                 encountered_errors[validation_result.row] = validation_result.errors
             else:
+                import_ids_to_create.append(validation_result.import_id)
                 document_create_objects.append(validation_result.create_request)
```
joel-wright, line 217: we could be skipping the import ids that already exist here, rather than in the background task...

backend/app/api/api_v1/routers/admin.py
```diff
@@ -276,6 +253,43 @@ async def import_law_policy(
         ) from e
 
 
+def start_import(db, s3_client, document_create_objects):
+    document_parser_inputs: list[DocumentParserInput] = []
+    documents_ids_already_exist: list[str] = []
```
joel-wright, line 258: we don't need to calculate this again, and the return value won't be used anyway

backend/app/api/api_v1/routers/admin.py
diversemix: Let me do that

backend/app/api/api_v1/routers/admin.py
diversemix: I didn't check this - thanks :face_exhaling:

joel-wright (approved):
lgtm

PR #854, opened 2022-11-03
Update endpoints to include slug and import_id

diversemix:


joel-wright (approved):
Once that `json_serialise` function is moved into test utils this looks good to go :)

joel-wright (approved):

PR #856, opened 2022-11-07
BAK-1240 Stage 2: Add postfix to document json (BACKEND)

diversemix:
Change everywhere a document json representation is used to include `postfix` - also updates tests.


backend/app/db/crud/document.py
```diff
@@ -370,6 +371,7 @@ def get_document_detail(db, import_id_or_slug) -> DocumentDetailResponse:
     return DocumentDetailResponse(
         id=document_id,
         name=cast(str, document.name),
+        postfix=cast(str, document.postfix),
```
joel-wright, line 374: Looking at this, I think we need to change `postfix` & `source_url` to be `Optional[str]` here, or we're likely to have a bad time. The old cast of `source_url` is from a naive time when we thought all documents required a source :(

backend/tests/routes/test_documents.py
```diff
@@ -199,6 +202,7 @@ def create_4_documents(test_db, client, superuser_token_headers):
     document4_payload = {
         "publication_ts": "1997-01-01T00:00:00.000000+00:00",
         "name": "Energy Sector Strategy 1387-1391 (2010/8-2015/3)",
+        "postfix": "postfix4",
```
joel-wright, line 205: We do probably also need to test out the option of `postfix=None`

joel-wright (approved):
A couple of comments re. typing & testing. Also, I assume there's a follow-up for adding appropriate postfixes to search results?

backend/tests/routes/test_documents.py
diversemix: good point

diversemix:
> A couple of comments re. typing & testing. Also, I assume there's a follow-up for adding appropriate postfixes to search results?

Yeah see the ticket for the stages: https://linear.app/climate-policy-radar/issue/BAK-1240/static-mapping-of-extra-info-to-add-to-title

PR #871, opened 2022-11-09
Show the geography name in search results

diversemix:
This PR creates, `getCountryName` and `getCountrySlug` both as a function of the country code

### DIFF
--- a/frontend/src/components/document/DocumentListItem.tsx
+++ b/frontend/src/components/document/DocumentListItem.tsx
@@ -5,6 +5,8 @@ import { truncateString } from "@helpers/index";
 import { getCategoryIcon } from "@helpers/getCatgeoryIcon";
 import { CountryLink } from "@components/CountryLink";
 import { ThemeContext } from "@context/ThemeContext";
+import { getCountryName } from "@helpers/getCountryFields";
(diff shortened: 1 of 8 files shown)


frontend/src/components/document/DocumentListItem.tsx
```diff
@@ -19,9 +21,12 @@ type TProps = {
 };
 
 export const DocumentListItem: FC<TProps> = ({ children, listItem }) => {
-  const { slug, country_code, country_name, description, name, document_year, category } = listItem;
+  const { slug, country_code, description, name, document_year, category } = listItem;
```
diversemix, line 24: @PatFawbertMills - I've taken country_name out of here - should I 
1. Update the components that use this to remove the country_name as well - or - 
2. Move the responsibility of finding the name into the parent compoents?

PatFawbertMills: More inclined to go with 1 - this is more elegant in my opinion. I don't tend to like overloading parent components with concerns of their children.

diversemix: done

PatFawbertMills (approved):
Good one

PR #889, opened 2022-11-09
change groupby and title sort key to document_name_and_slug

kdutia:
once the new indexing run is done this should fix sort by title


diversemix (approved):
lgtm

PR #896, opened 2022-11-10
Re-enable search tests

kdutia:
* new opensearch data dump containing `document_sectors` (see https://github.com/climatepolicyradar/navigator-search-indexer/pull/35)
* unskipped tests and fixed 
* better `test_browse_limit_offset` test


backend/app/core/browse.py
```diff
@@ -32,6 +32,7 @@ def to_search_resp_doc(row: dict) -> SearchResult:
         document_category=row["category"],
         document_geography=row["country_code"],
         # Stuff we don't use
+        document_sectors=[],  # FIXME: empty placeholder for tests to pass
```
kdutia, line 35: I wasn't sure how to deal with many sectors per document in RDS so have left this as a `FIXME`. Happy to be guided through it or for one of you to do it if it's quick :)

backend/tests/routes/test_search.py
```diff
@@ -20,7 +20,6 @@
 _TOTAL_DOCUMENT_COUNT = 6
 
 
-@pytest.mark.skip(reason="new test data dump needed")
 @pytest.mark.search
 def test_simple_pagination(test_opensearch, monkeypatch, client):
```
diversemix, line 24: Do we still not have enough data to at least get 2 items on a page?

kdutia: we do - i can update the test

backend/tests/routes/test_search.py
```diff
@@ -129,7 +127,7 @@ def test_pagination_overlap(test_opensearch, monkeypatch, client):
 
     page1_response_body = page1_response.json()
     page1_documents = page1_response_body["documents"]
-    assert len(page1_documents) == 2
+    assert len(page1_documents) > 0
```
diversemix, line 130: To test the overlap - we probably need `>1`

backend/app/core/browse.py
joel-wright: Ah yeah, document sectors will need to be created separately, I'll have a look at this.

backend/tests/routes/test_search.py
```diff
@@ -30,42 +27,41 @@ def test_simple_pagination(test_opensearch, monkeypatch, client):
...
         },
     )
     assert page2_response.status_code == 200
 
     page2_response_body = page2_response.json()
     page2_documents = page2_response_body["documents"]
-    assert len(page2_documents) == 1
+    assert len(page2_documents) == 2
 
     # Sanity check that we really do have 2 different documents
```
joel-wright, line 55: :grin: 

```suggestion
    # Sanity check that we really do have 4 different documents
```

joel-wright (approved):
LGTM

backend/app/core/browse.py
joel-wright: We can do this afterwards though, no need to hold up the merge for this.

PR #902, opened 2022-11-15
BAK-1260: Test bash scripts

diversemix:
This PR is designed to introduce some testing for the bash script that does the retagging and pushing to ECR

- Adds a docker with bats support
- Rejig some of the code so its testable
- Create a first useful test


henry-cpr (approved):
Looks great! The main bit that made me anxious was pulling out the major/minor/etc components of the semver string, so I think if we can get that covered we'll have a nice foundation

diversemix:
Yup that should all be done now @henry-cpr

PR #906, opened 2022-11-24
Adobe key into the context and take from server in _app

PatFawbertMills:
Ensuring we are able to get the adobe key from the server environment variables before rendering the Adobe PDF viewer

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -63,6 +63,9 @@ services:
     stdin_open: true
     environment:
       NODE_ENV: development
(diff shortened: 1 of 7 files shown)


joel-wright (approved):
As we talked about on Friday, this is phase 1 of fixing the environment issue. Phase 2 will come in a separate PR & will focus on fixing the redirects using a middleware. When that is complete we should be able to deploy a single front-end image using only environment variables to specify CCLW/CPR configuration....

PR #907, opened 2022-11-27
WIP

joel-wright:


frontend/src/types/index.ts
```diff
@@ -16,6 +16,12 @@ export type TSector = {
   source_id: number;
 };
 
+export type TRedirect = {
```
joel-wright, line 19: :+1:

frontend/src/middleware.mts
```diff
@@ -13,43 +13,37 @@ const REDIRECT_FILE = process.env.NEXT_REDIRECT_FILE || "default.csv";
...
-    let redirect_map = new Map<string, object>();
-    standardRedirects.forEach(redirect => {
-        redirect_map.set(redirect["source"], redirect)
-    });
-
-    return redirect_map
+    return standardRedirects.reduce((acc, item) => {
+        acc[item["source"]] = item;
+        return acc;
+    }, new Map<string, TRedirect>());
```
joel-wright, line 36: This is nice (even if reading the docs for `reduce` did make me sad)

frontend/next.config.js
```diff
@@ -2,7 +2,7 @@
   THIS IS BURNED AT BUILDTIME - ENV ARE IGNORED AT RUNTIME
 */
```
joel-wright, line 4: I don't know if it's important (and I deleted some of it by being a touch over-eager), but all the next config examples look like this:
```javascript
/**
 * @type {import('next').NextConfig}
 */
const nextConfig = {
  /* config options here */
}

module.exports = nextConfig
```

joel-wright (commented):
Thanks for making those changes, have made comments in an attempt to properly understand...

frontend/next.config.js
diversemix: I think you are right - let me put this in :+1:

frontend/src/middleware.mts
```diff
@@ -0,0 +1,49 @@
...
+    { source: "/contact", destination: "/", permanent: true },
+    { source: "/methodology", destination: "/", permanent: true },
+];
+
+async function get_redirects(): Promise<Map<string, TRedirect>> {
+    let standardRedirects = process.env.THEME === "cclw" ? cclwRedirects : cprRedirects;
+    standardRedirects.concat(await read(REDIRECT_FILE));
+
+    return standardRedirects.reduce(
+        (acc, item) => ( acc.set(item["source"], item) ),
```
diversemix, line 34: The previous return statement here was bugging me - so I removed it to simplify.

diversemix (approved):
Joint approve with @joel-wright

joel-wright:
Thanks for fixing my shonky javascript - happy to merge when you think it's okay

frontend/next.config.js
```diff
@@ -0,0 +1,70 @@
...
+    {
+      source: "/litigation/:id*",
+      destination: "/",
+      permanent: false, // will become a page eventually
+    },
+  ];
+ 
+
+function get_redirects() {
+    const REDIRECT_FILE = process.env.NEXT_REDIRECT_FILE || "default.json";
```
PatFawbertMills, line 40: I think this will only ever return the default.json because from my understanding the config is only executed at build time

PR #909, opened 2022-11-30
FRO 129: Upgrade dependencies as a first step

diversemix:
It was noted that in order to get a stable version of middleware and to be able to use the config from a function, we needed to update Nextjs. Was decided to go with the latest version, this also enforced React v18 ... so that was done as well.
- upgrade next and react
- audit fix vulnerable deps
- upgrade `react-i18next`
- run `codemod` as suggested by Nextjs
- manually fix linting issues - mainly adding a prop attribute of `children`


frontend/package.json
```diff
@@ -15,20 +15,20 @@
...
+    "babel-loader": "^9.1.0",
     "cypress-pseudo-localization": "^1.0.0",
     "formik": "^2.2.9",
     "i18next": "^21.6.11",
     "i18next-browser-languagedetector": "^6.1.3",
     "i18next-http-backend": "^1.3.2",
-    "next": "latest",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2",
+    "next": "^13.0.0",
```
joel-wright, line 24: this is better :)

joel-wright (commented):
Seems reasonable to me

frontend/src/components/CountryLink.tsx
```diff
@@ -1,11 +1,12 @@
-import { FC } from "react";
+import { FC, ReactNode } from "react";
 import Link from "next/link";
 import useConfig from "@hooks/useConfig";
 import { getCountrySlug } from "@helpers/getCountryFields";
 
 type TCountryLink = {
   countryCode: string;
   className?: string;
+  children?: ReactNode;
```
PatFawbertMills, line 9: I don't think we need to declare the children type on the TCountryLink if we are inheriting the FC (functional component) type

frontend/src/components/CountryLink.tsx
```diff
@@ -15,8 +16,10 @@ export const CountryLink: FC<TCountryLink> = ({ countryCode, className = "", chi
   const slug = getCountrySlug(countryCode, countries);
   if (!slug) return <>{children}</>;
   return (
-    <Link href={`/geographies/${slug}`}>
-      <a className={`flex items-center underline ${className}`}>{children}</a>
+    <Link
+      href={`/geographies/${slug}`}
+      className={`flex items-center underline ${className}`}>
```
PatFawbertMills, line 21: I am getting an error because the property 'className' does not exist on the typings for 'Link'

PatFawbertMills (approved):
Looks good, nice work Peter.

PR #910, opened 2022-11-30
FRO-129 Fix redirects

diversemix:
Current state of this branch is that it works - but it does load the redirects for every incoming request.


PatFawbertMills (approved):
Looks good with middleware

PR #914, opened 2022-12-09
Feature/fro 146 display number of results

PatFawbertMills:
We now display the number of results that have been returned. And will display an info tooltip if the results are capped to 100.

Screenshot 2022-12-09 at 14 03 13


diversemix (approved):
Nice. TIL `passHref`

PR #917, opened 2022-12-15
Feature/robots txt sitemap

PatFawbertMills:
Requires two new env vars to be created:
- ROBOTS: this needs to be a string of "true" in order to have an effect. It will allow the site to be crawled.
- HOSTNAME: this is to be used when creating the sitemap

Sitemap is now generating the geography pages using their slug rather than ID.


diversemix (approved):
Looks good - also opened https://linear.app/climate-policy-radar/issue/BAK-1282/update-infra-to-use-new-vars-in-apprunner

PR #921, opened 2022-12-20
Configure AWS credentials for SemVer Tagging

joel-wright:


THOR300 (approved):
Only thing I would check is whether region is listed as a secret as well and use that.

joel-wright:
> Only thing I would check is whether region is listed as a secret as well and use that.

Yeah, I did think about it, but knowing the region seems safe, and we have this pattern elsewhere...

PR #922, opened 2022-12-22
Rewrite landing page animation

PatFawbertMills:
I’ve re-written the “typewriter” animation on the homepage to use css animations rather than manipulating the input’s placeholder as it was before. This means we will see the end of that annoying flakey front-end test that fails due to timing issues when running in CI.

I’ve also written some much better e-2-e tests on this element to test that the animated element hides when the user starts typing, and that what they type is being logged. Just a start, but feels to have that 1% more coverage.

### DIFF
--- a/frontend/e2e/cypress/e2e/landing/landing-page.spec.js
+++ b/frontend/e2e/cypress/e2e/landing/landing-page.spec.js
@@ -5,20 +5,31 @@ describe("Landing page", () => {
     cy.visit("/");
   });
 
-  it("should see placeholder text which should clear after clicking input", () => {
-    // wait for text animation to finish
-    cy.wait(3000);
-
+  it("should see placeholder text", () => {
     cy.get('[data-cy="search-input"]')
       .invoke("attr", "placeholder")
       .then((text) => {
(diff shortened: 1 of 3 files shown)


frontend/e2e/cypress/e2e/landing/landing-page.spec.js
```diff
@@ -5,20 +5,31 @@ describe("Landing page", () => {
     cy.visit("/");
   });
 
-  it("should see placeholder text which should clear after clicking input", () => {
-    // wait for text animation to finish
-    cy.wait(3000);
-
```
joel-wright, line 11: this makes me so damned happy :joy:

joel-wright (approved):
I don't understand the CSS, but what you sent earlier looked great & I love that we don't have horrible test sleeps!!

frontend/e2e/cypress/e2e/landing/landing-page.spec.js
PatFawbertMills: Yeh 3 seconds off every run feels good

PR #925, opened 2023-01-03
Add some debugging to determine the issue

diversemix:


THOR300 (approved):
LGTM - added some additional logs as well

PR #926, opened 2023-01-04
Re-enable default prefetch on links

PatFawbertMills:
Prefetch was removed when we ran into issues with caching.

We've now looked into it further and there is no need to remove it.


diversemix (approved):
lgtm

PR #930, opened 2023-01-07
Allow all acceptable origins via regex

joel-wright:


THOR300 (approved):
Looks good, from my understanding looks like the backend instances would be able to accept requests from any domain name. I.e. dev backend could be pinged by sandbox domain name if we're okay with that.

joel-wright:
> Looks good, from my understanding looks like the backend instances would be able to accept requests from any domain name. I.e. dev backend could be pinged by sandbox domain name if we're okay with that. 

It's a fair point, but it is what we currently do & having the flexibility has proved useful. We might want to look into making this configurable at deploy time in the future...

diversemix:
This could be easily done with CORS

PR #931, opened 2023-01-08
Remove nginx references

joel-wright:


.env.example
```diff
@@ -71,3 +71,5 @@ CDN_DOMAIN=cdn.climatepolicyradar.org
 # Pipeline settings
 PIPELINE_BUCKET="a-bucket-for-testing-cpr-endpoints"
 
+# Default to CPR theme
+THEME=cpr
```
diversemix, line 75: May as well add ROBOTS, and HOSTNAME as well

nginx/nginx.conf
```diff
@@ -1,36 +0,0 @@
...
-        proxy_pass http://backend:8888/api/tokens;
-    }
-
-    location /health {
-        proxy_pass http://backend:8888/health;
-    }
-
-    location ~ ^/test/(.*)$ {
-        resolver 8.8.8.8;
-        proxy_pass https://httpbin.org/$1$is_args$args;
```
diversemix, line 34: @PatFawbertMills are we good to remove this?

diversemix (changes requested):
lgtm

diversemix (approved):
lgtm

.env.example
joel-wright: Will do this as a follow-up

PR #936, opened 2023-01-11
Fix sitemap rendering by removing trailing slash on config endpoint

diversemix:


frontend/.env.example
```diff
@@ -6,4 +6,4 @@ LOGIN_API_URL=http://localhost:8000/api/tokens
 THEME=cpr
 
 # Front-end redirects
-NEXT_REDIRECT_FILE="default.csv"
+NEXT_REDIRECT_FILE="default.json"
```
diversemix, line 9: The old `default.csv` has also been removed to avoid confusion - redirects must be specified in a `JSON` file

joel-wright (approved):

